PR #85, opened 2019-02-19
Refactor and  move some code utils to common codebase

zguesmi:
- Consider tx failed when txReceipt is null
- Move notification related classes to their own package.
- Move signature & eip712Challenge utils to the common codebase.


james-toussaint (approved):
Nice clean up!

Ugo (approved):

PR #86, opened 2019-02-20
Add sgx in the configuration

Ugo:


james-toussaint (changes requested):
We shouldn't use SGX, we have to use 'TEE' everywhere

Ugo:
Right, changed.

james-toussaint (approved):
Thank you!

PR #89, opened 2019-03-01
Move web3 and hub to common

james-toussaint:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -0,0 +1,252 @@
...
+                    .checksum(BytesUtils.bytesToString(app.m_appChecksum().send()))
+                    .build());
+        } catch (Exception e) {
+            log.error("Failed to get ChainApp [chainAppId:{}]", app.getContractAddress());
+        }
+        return Optional.empty();
+    }
+
+    public Optional<ChainDataset> getChainDataset(Dataset dataset) {
+        if (!dataset.getContractAddress().equals(BytesUtils.EMPTY_ADDRESS)) {
```
zguesmi: We should handle cases where dataset or dataset.getContractAddress() are null

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -0,0 +1,252 @@
...
+                    .schedulerRewardRatio(config.getValue6())
+                    .build());
+        } catch (Exception e) {
+            log.error("Failed to get ChainDeal [chainDealId:{}]", chainDealId);
+        }
+        return Optional.empty();
+    }
+
+    public Optional<ChainTask> getChainTask(String chainTaskId) {
+        ChainTask chainTask = null;
```
zguesmi: maybe we can remove this since it's not used

src/main/java/com/iexec/common/chain/Web3jAbstractService.java
```diff
@@ -0,0 +1,233 @@
...
+    public Web3j getWeb3j() {
+        return web3j;
+    }
+
+    private EthBlock.Block getLatestBlock() throws IOException {
+        return web3j.ethGetBlockByNumber(DefaultBlockParameterName.LATEST, false).send().getBlock();
+    }
+
+    private long getLatestBlockNumber() throws IOException {
+        return getLatestBlock().getNumber().longValue();
```
zguesmi, line 62: same here, getLatestBlock() can be null (?)

Ugo (approved):

PR #90, opened 2019-03-04
Pending receipt

james-toussaint:


Ugo (approved):
Much cleaner with default methods. Thanks!

PR #98, opened 2019-03-15
Send full params field when params is not a JSON (or invalid)

james-toussaint:


src/main/java/com/iexec/common/chain/ChainDeal.java
```diff
@@ -45,16 +46,15 @@
...
-    public static List<String> stringParamsToList(String params){
-        List<String> listParams = new ArrayList<>();
+    public static List<String> stringParamsToList(String params) {
+        List<String> listParams;
         try {
             LinkedHashMap tasksParamsMap = new ObjectMapper().readValue(params, LinkedHashMap.class);
             listParams = new ArrayList<String>(tasksParamsMap.values());
         } catch (IOException e) {
-            e.printStackTrace();
+            listParams = Collections.singletonList(params);//the requester want to execute one task with the whole string
```
Ugo, line 58: Maybe it would be good to have a log here.

Ugo (approved):

src/main/java/com/iexec/common/chain/ChainDeal.java
james-toussaint: I think its going to happen very frequently, it allows the requester sending one task (very usual) to only send `bla` instead of `{ "0" : "bla" }`.
Do you think we still need it?

PR #102, opened 2019-03-26
Jenkins

james-toussaint:


build.gradle
```diff
@@ -55,14 +62,14 @@ shadowJar { //required if we want to embed web3j dependency inside
 build.dependsOn(shadowJar)
 
 def gitBranch = 'git name-rev --name-only HEAD'.execute().text.trim()
-def isMasterBranch = gitBranch == "remotes/origin/master"
+def isMasterBranch = true//gitBranch == "remotes/origin/master"
```
Ugo: isMasterBranch is not always true :-)

Ugo (changes requested):

build.gradle
james-toussaint: Thank you!

Ugo (approved):

Ugo (approved):

PR #122, opened 2019-05-24
Add block number to notifications

Ugo:


james-toussaint (commented):
Great, I would had `transactionHash` too

james-toussaint (approved):

PR #126, opened 2019-06-10
Wait ethereum node if missing at startup

james-toussaint:


Ugo (approved):
Simple solution, perfect.

PR #127, opened 2019-06-12
Add method to get the consensus reached block for a task + isBlockAvailable fix

Ugo:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -350,4 +352,20 @@ protected boolean isStatusValidOnChainAfterPendingReceipt(String chainTaskId, Ch
                 .build());
     }
 
+    public long getConsensusReachedBlock(String chainTaskId, long fromBlock) {
```
james-toussaint: Cool !
I would just check if endBlock > startBlock (?)

Ugo: yes, added

james-toussaint (approved):
Great feature, thanks :)

PR #128, opened 2019-06-13
Add details tools

Ugo:


james-toussaint (approved):
Great!
Maybe we can have getConsensusBlock(), getConsensusBlock().getNumber(),  getConsensusBlock().getTxNumber(), etc so we can jump to an etherscan, blockscout explorer

PR #131, opened 2019-06-21
Implement full TEE with sgx

zguesmi:


src/main/java/com/iexec/common/replicate/ReplicateStatus.java
```diff
@@ -20,6 +20,9 @@
     COMPUTED,
     COMPUTE_FAILED,
     CAN_CONTRIBUTE,
+    CANT_CONTRIBUTE_SINCE_DETERMINISM_HASH_NOT_FOUND,
```
james-toussaint, line 23: We should update getUncompletableStatuses() and co

src/main/java/com/iexec/common/utils/SignatureUtils.java
```diff
@@ -39,6 +40,27 @@ public static boolean doesSignatureMatchesAddress(byte[] signatureR,
         return false;
     }
 
+    public static boolean isSignatureValid(String message, Signature sign, String signerAddress) {
+        return isSignatureValid(message.getBytes(), sign, signerAddress);
+    }
+
+    public static boolean isSignatureValid(byte[] message, Signature sign, String signerAddress) {
```
Ugo, line 43: This is an important method here, good unit tests are mandatory.

Ugo (changes requested):

src/main/java/com/iexec/common/utils/SignatureUtils.java
zguesmi: Sure yes! thank you for the note.

src/main/java/com/iexec/common/replicate/ReplicateStatus.java
zguesmi: Done, thank you!

james-toussaint (approved):

Ugo (approved):

Ugo:
Thank you Zied!

PR #136, opened 2019-07-16
Replicate status cause

james-toussaint:


Ugo (approved):
Easier to read now, thanks

PR #137, opened 2019-07-18
Modify chainDeal params

Ugo:


src/main/java/com/iexec/common/chain/DealParams.java
```diff
@@ -0,0 +1,19 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+@Getter
+public class DealParams {
+
+    // Note to dev: the naming of the variables here is important since it will be stored on-chain
+    // that is why the names are not java compliant (use of underscore and no upper case)
+    private String iexec_args;
+    private List<String> iexec_input_files;
```
zguesmi: You can use iexecArgs and iexecInputFiles, the mapping is done automatically by ObjectMapper.

src/main/java/com/iexec/common/chain/ChainDeal.java
```diff
@@ -48,16 +48,20 @@
...
         try {
-            LinkedHashMap tasksParamsMap = new ObjectMapper().readValue(params, LinkedHashMap.class);
-            listParams = new ArrayList<String>(tasksParamsMap.values());
+            DealParams dealParams = new ObjectMapper().readValue(params, DealParams.class);
+            if(dealParams.getIexec_input_files() == null) {
+                dealParams.setIexec_input_files(new ArrayList<>());
+            }
+            return dealParams;
         } catch (IOException e) {
             log.warn("Params string is not a JSON, considering the string is one full param");
```
james-toussaint: You can remove this log (to many occurrences in prod)

james-toussaint (approved):

src/main/java/com/iexec/common/chain/DealParams.java
Ugo: The mapping is not working if iexecArgs and iexecInputFiles are used.

zguesmi: I think the @JsonProperty is needed.

Ugo: ok good with the proper JsonProperty

src/main/java/com/iexec/common/chain/ChainDeal.java
Ugo: done

PR #149, opened 2019-09-02
Add more accurate failure status causes

zguesmi:


Ugo (approved):

src/main/java/com/iexec/common/replicate/ReplicateActionResponse.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.NoArgsConstructor;
+
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class ReplicateActionResponse {
+
+    private ReplicateStatus status;
```
james-toussaint: Don't we want a boolean? (Sorrry)

src/main/java/com/iexec/common/replicate/ReplicateActionResponse.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class ReplicateActionResponse {
```
james-toussaint, line 17: Great!

src/main/java/com/iexec/common/replicate/ReplicateActionResponse.java
zguesmi: No we don't need it since we'll do the same thing independently from whether it is a fail or a success.

Ugo (approved):

PR #165, opened 2020-02-20
Goerli fix

james-toussaint:


src/main/java/com/iexec/common/chain/Web3jAbstractService.java
```diff
@@ -21,48 +21,54 @@
...
-                return web3j;
+    public Web3j getWeb3j(boolean shouldCheckConnection) {
+        web3j = Web3j.build(new HttpService(chainNodeAddress));
+        if (shouldCheckConnection){
+            try {
+                if (web3j.web3ClientVersion().send().getWeb3ClientVersion() != null) {
+                    log.info("Connected to Ethereum node [address:{}, version:{}]", chainNodeAddress, web3j.web3ClientVersion().send().getWeb3ClientVersion());
+                    return web3j;
+                }
+            } catch (IOException ignored) {
```
zguesmi: Maybe we should log the exception (or at least the exception message) here. It could be useful (for exp when web3j is null).

src/main/java/com/iexec/common/chain/Web3jAbstractService.java
```diff
@@ -21,48 +21,54 @@
...
             }
-        } catch (IOException ignored) {
+
+            int fewSeconds = 5;
+            WaitUtils.sleep(fewSeconds);
+            log.error("Failed to connect to ethereum node (will retry) [chainNodeAddress:{}, retryIn:{}]",
```
zguesmi, line 58: I think it's better to invert the two lines above to log then sleep.

src/main/java/com/iexec/common/chain/Web3jAbstractService.java
james-toussaint: Thanks !

PR #166, opened 2020-02-21
Added getAverageTimePerBlock() & repeatCheck() methods

james-toussaint:


zguesmi (approved):
Code that runs code, so sexy!

PR #178, opened 2020-04-20
Sign inside service without private key disclosure & clean old sms legacy

zguesmi:


src/main/java/com/iexec/common/chain/CredentialsAbstractService.java
```diff
@@ -42,4 +46,9 @@ public CredentialsAbstractService(String walletPassword, String walletPath) thro
     public Credentials getCredentials() {
         return credentials;
     }
+
+    public Signature hashAndSignMessage(String message) {
```
james-toussaint, line 53:
```suggestion
    /* 
    * Signs messages with Ethereum prefix
    */
    public Signature hashAndSignMessage(String message) {
```

james-toussaint (approved):
Excellent cleaning, I love that!

PR #179, opened 2020-04-22
Moved post-compute encryption helper to common, required in integration

james-toussaint:


src/main/java/com/iexec/common/security/EncryptionHelper.java
```diff
@@ -0,0 +1,179 @@
...
+
+import static com.iexec.common.security.CipherHelper.*;
+import static com.iexec.common.utils.FileHelper.*;
+
+@Slf4j
+public class EncryptionHelper {
+
+    public static final String AES_KEY_RSA_FILENAME = "aes-key.rsa";
+    public static final String ENCRYPTION_PREFIX = "encrypted-";
+    public static final String DECRYPTION_PREFIX = "clear-";
```
zguesmi: Can we use "plain" instead of "clear" since it's more used and intuitive?

zguesmi:
```suggestion
    public static final String DECRYPTION_PREFIX = "plain-";
```

zguesmi (approved):

PR #184, opened 2020-04-30
Move static config to common

zguesmi:


src/main/java/com/iexec/common/sms/SmsConfig.java
```diff
@@ -0,0 +1,6 @@
+package com.iexec.common.sms;
+
+public class SmsConfig {
+
+    public static final String DOMAIN = "IEXEC_SMS_DOMAIN"; // TODO: Add session salt after domain
```
james-toussaint: DOMAIN should depends on SMS version, e.g IEXEC_SMS_DOMAIN_v1.0.0, I don't think we can move it to common (or we can call it DOMAIN_PREFIX)

zguesmi: Renaming it does not help, so I'll leave it in the SMS.

james-toussaint (approved):

PR #185, opened 2020-05-15
Feature/computed json

james-toussaint:


src/main/java/com/iexec/common/result/ComputedFile.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ComputedFile {
+
+    @JsonProperty("deterministic-output")
```
zguesmi: Maybe renaming it to `deterministic-output-path` would be clearer and more explicit.

zguesmi (approved):

src/main/java/com/iexec/common/result/ComputedFile.java
james-toussaint: Sure, but developer are going to write it down pretty often, so we need to decide between short & clear

zguesmi: Good point, personally I'd choose clearness since it might cause execution to fail in case the developer doesn't understand the meaning of the attribute.

james-toussaint: Updated, thanks

zguesmi (approved):

zguesmi:
Thanks !

PR #193, opened 2020-06-03
Add stdout to status update details

zguesmi:


src/main/java/com/iexec/common/replicate/ReplicateStatusDetails.java
```diff
@@ -25,4 +37,11 @@ public ReplicateStatusDetails(long blockNumber) {
     public ReplicateStatusDetails(ReplicateStatusCause cause) {
         this.cause = cause;
     }
+
+    public ReplicateStatusDetails cropStdout() {
```
james-toussaint: I would `tail stdout` instead of `head stdout`

zguesmi: Thanks for noticing it !

james-toussaint (approved):

PR #205, opened 2020-10-16
Read onchain constant only once

zguesmi:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -84,6 +85,14 @@ public IexecHubAbstractService(Credentials credentials,
 
         String hubAddress = getHubContract().getContractAddress();
         log.info("Abstract IexecHubService initialized (iexec proxy address) [hubAddress:{}]", hubAddress);
+
+        try {
+            this.maxNbOfPeriodsForConsensus = getHubContract()
+                    .contribution_deadline_ratio().send().longValue();
+        } catch (Exception e) {
+            log.error("Failed to get maxNbOfPeriodsForConsensus from the chain");
+            System.exit(1);
```
james-toussaint: Why do we exit?

zguesmi: Sorry I'll fix that.

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -354,12 +356,20 @@ public String getOwner(String address) {
...
+
+    private void setMaxNbOfPeriodsForConsensus() {
         try {
-            return getHubContract().contribution_deadline_ratio().send().longValue();
+            this.maxNbOfPeriodsForConsensus = getHubContract()
+                    .contribution_deadline_ratio().send().longValue();
         } catch (Exception e) {
-            log.error("Failed to getMaxNbOfPeriodsForConsensus");
+            log.error("Failed to get maxNbOfPeriodsForConsensus from the chain");
+            this.maxNbOfPeriodsForConsensus = 0;
```
james-toussaint: Don't you think -1 is more appropriate for error return (since 0 could be a real value, at least in other contexts) ?

zguesmi: This value is used to determine the TTL of an executor thread which cannot be a negative value and is also used to calculate dates which also cannot be negative. At the beginning I used exit() because of that but using 0 as fallback can be a better alternative since it gives a valid value that means the task is expired.

james-toussaint (approved):

PR #213, opened 2020-12-15
Get restricted orders

zguesmi:


src/main/java/com/iexec/common/sdk/marketplace/Marketplace.java
```diff
@@ -33,9 +33,13 @@
 
     public Optional<AppMarketOrder> getOneAppOrder(String appAddress);
 
+    public Optional<AppMarketOrder> getOneRequesterRestrictedAppOrder(String appAddress, String requesterAddress);
```
james-toussaint, line 36: What about `getOneAppOrder(String appAddress, String requesterAddress)`?

james-toussaint (approved):

src/main/java/com/iexec/common/sdk/marketplace/Marketplace.java
zguesmi: Yeah, I thought about that, but if we want to add `getAppOrder(String appAddress, String appOwner)` it wouldn't be possible.

james-toussaint: :+1:

PR #223, opened 2021-02-05
Add one common docker package

zguesmi:


src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.iexec.common.docker.client;
+
+public abstract class DockerClientFactory {
+
+    private static DockerClientInstance defaultClient;
+    private static DockerClientInstance authenticatedClient;
```
james-toussaint: A map is probably more appropriate for auth clients

james-toussaint (approved):

src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
zguesmi: My bad, ugly copy-paste. Sorry!

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -0,0 +1,663 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * Docker container
+     * 
+     */
+
+    public DockerRunResponse run(DockerRunRequest dockerRunRequest) {
```
james-toussaint, line 313: Will be hard to test

zguesmi: I agree but we don't have a choice for now.
PS: tests coming in the next PR.

james-toussaint (approved):

PR #229, opened 2021-03-19
Decrypt dataset in prec-ompute

zguesmi:
* Upgrade gradle to 6.8.3
* Add pre-compute utils


james-toussaint (approved):

src/main/java/com/iexec/common/security/CipherUtils.java
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    /**
+     * Remove prepended IV (initialization vector)
+     * from the encrypted data.
+     * 
+     * @param dataWithIv
+     * @return a new array without the IV.
+     */
+    public static byte[] stripIvFromEncryptedData(byte[] dataWithIv) {
```
james-toussaint, line 116: Maybe soemething like
`getEncryptedDataWithoutIv(..)`
since `stripIvFromEncryptedData(..)` lets think original array will be updated

src/main/java/com/iexec/common/security/CipherUtils.java
```diff
@@ -0,0 +1,384 @@
...
+    /**
+     * Concatenate IV (initialization vector) to the
+     * beginning of the encrypted data.
+     * 
+     * @param iv
+     * @param encryptedData
+     * @return
+     * @throws IOException
+     */
+    public static byte[] prependIvToEncryptedData(byte[] iv, byte[] encryptedData)
```
james-toussaint, line 129: concatenateIvAndEncryptedData(..)
?

src/main/java/com/iexec/common/security/CipherUtils.java
```diff
@@ -0,0 +1,384 @@
...
+     * @param encryptedDataWithIv to decrypt
+     * @param base64Key Base64 encoded AES key
+     * @return Decrypted data
+     * @throws GeneralSecurityException
+     * @throws IllegalArgumentException if the data's length
+     * is less than 16 bytes.
+     * 
+     * @see https://stackoverflow.com/a/34004582 for large files
+     */
+    public static byte[] aesDecrypt(byte[] encryptedDataWithIv, byte[] base64Key)
```
james-toussaint: byte[] ivAndEncryptedData
?

james-toussaint (approved):
Great work!

src/main/java/com/iexec/common/security/CipherUtils.java
zguesmi: I don't think `strip` does necessarily mean that we are modifying the array itself. `String#strip()` for example does not do that. I can still rename it if you still think it can be confusing.

src/main/java/com/iexec/common/security/CipherUtils.java
zguesmi: Concatenate does not give an idea about the order, while `prepend` tells explicitly that the IV will be at the beginning.

src/main/java/com/iexec/common/security/CipherUtils.java
zguesmi: Yes, it makes more sense.

PR #235, opened 2021-03-31
Feature/dataset filename for standard compute stage

james-toussaint:


src/main/java/com/iexec/common/utils/FileHelper.java
```diff
@@ -131,6 +134,11 @@ public static String downloadFile(String fileUri, String downloadDirectoryPath)
             return "";
         }
 
+        if (outputFilename.isEmpty()) {
```
zguesmi:
```suggestion
        if (StringUtils.isBlank(outputFilename)) {
```
Since it can be null.

zguesmi: Also, let's move it to the beginning of the method so we fail faster before creating the folder.

zguesmi (approved):

PR #236, opened 2021-04-02
Ad pre-compute in ou out properties

zguesmi:


src/main/java/com/iexec/common/utils/IexecFileHelper.java
```diff
@@ -26,6 +26,11 @@
 @Slf4j
 public class IexecFileHelper {
 
+    public static final String SLASH_PRE_COMPUTE = File.separator + "pre-compute";
```
james-toussaint: Is that an in or out path?

james-toussaint (approved):

src/main/java/com/iexec/common/utils/IexecFileHelper.java
zguesmi: I'll append `-in` to the path.

PR #239, opened 2021-05-18
Added pre-compute config

james-toussaint:


src/main/java/com/iexec/common/precompute/PreComputeConfig.java
```diff
@@ -0,0 +1,29 @@
...
+    public PreComputeConfig(String image, String fingerprint, String heapSize) {
+        if (StringUtils.isEmpty(image)) {
+            throw new IllegalStateException("No pre-compute image is provided");
+        }
+        if (StringUtils.isEmpty(fingerprint)) {
+            throw new IllegalStateException("No pre-compute fingerprint is provided");
+        }
+        if (StringUtils.isEmpty(heapSize)) {
+            throw new IllegalStateException("No pre-compute heap size is provided");
+        }
```
zguesmi: I think this part is better kept on the SMS's side ([here](https://github.com/iExecBlockchainComputing/iexec-sms/pull/49/files#diff-850c9f9cbbf0eaf6af1153790bc90d676d4a5cb45b8c08607d4736de04245e90R18)).

zguesmi (approved):

src/main/java/com/iexec/common/precompute/PreComputeConfig.java
james-toussaint: Thanks!

PR #242, opened 2021-05-28
Fix mrenclave for standard tasks

zguesmi:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -484,23 +484,46 @@ public String getTaskResults(String chainTaskId, Integer chainId) {
...
+        if (StringUtils.isEmpty(mrEnclave)) {
+            // Standard application
+            return Optional.of(chainApp);
+        }
+        try {
+            chainApp.setEnclaveConfiguration(
+                    buildEnclaveConfigurationFromJsonString(mrEnclave));
+        } catch (Exception e) {
+            log.error("Failed to get tee chain app enclave configuration " +
+                    "[chainAppId:{}]", app.getContractAddress(), mrEnclave);
```
james-toussaint: MR enclave is missing from log
Exception is not logged

james-toussaint (approved):
Thanks!

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
zguesmi: Fixed, thanks!

PR #247, opened 2021-06-09
Add eth address utils

zguesmi:


src/main/java/com/iexec/common/utils/EthAddress.java
```diff
@@ -0,0 +1,85 @@
...
+     * Checks if the given string is a checksummed address. Inspired by
+     * the web3js implementation.
+     * 
+     * @param address in hex
+     * @return true if valid checksummed address, false otherwise
+     * @see https://github.com/ChainSafe/web3.js/blob/
+     * 5d027191c5cb7ffbcd44083528bdab19b4e14744/packages/web3-utils/src/
+     * utils.js#L110
+     */
+    public static boolean validateChecksummed(String address) {
```
james-toussaint: That one is not enough https://github.com/web3j/web3j/blob/master/crypto/src/main/java/org/web3j/crypto/Keys.java#L125?

zguesmi: It is indeed, I didn't know about it. Thank you!

james-toussaint (approved):

PR #248, opened 2021-06-14
Add authenticated docker clients

zguesmi:


src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
```diff
@@ -16,18 +16,71 @@
...
+    }
+
+    /**
+     * Used at least in unit tests.
+     */
+    static void purgeClients() {
+        clientsMap.clear();
+    }
+
+    private static DockerClientInstance createInstance(
```
james-toussaint: getOrCreate

src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
```diff
@@ -16,18 +16,71 @@
...
+    /**
+     * Get a Docker client instance that is authenticated against the default docker
+     * registry {@link AuthConfig#DEFAULT_SERVER_ADDRESS}
+     * ({@code https://index.docker.io/v1/}) using the provided credentials.
+     * 
+     * @param username
+     * @param password
+     * @throws DockerException if the authentication fails
+     */
+    public static synchronized DockerClientInstance getDockerClientInstance(
```
james-toussaint: Do we need this method? We might use explicit addresses to avoid confusions with "defaults"

src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
zguesmi: Updated, thanks!

src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
zguesmi: I agree we should treat the default docker registry as any other registry.

james-toussaint (approved):
Excellent!

PR #249, opened 2021-06-15
Refactor docker default config

zguesmi:


src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
```diff
@@ -47,7 +69,7 @@ public static synchronized DockerClientInstance getDockerClientInstance() {
      * @throws DockerException if the authentication fails
      */
     public static synchronized DockerClientInstance getDockerClientInstance(
-            String registryUrl, String username, String password) throws DockerException {
+            String registryUrl, String username, String password) throws Exception {
```
james-toussaint, line 69: Should check registryAddress?

src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
```diff
@@ -59,10 +81,15 @@ static void purgeClients() {
     }
 
     private static DockerClientInstance getOrCreateInstance(
-            String registryUrl, String username, String password) {
+            String registryUrl, String username, String password) throws Exception {
         String id = getClientIdentifier(registryUrl, username);
+        boolean isAuthenticated = StringUtils.isNotBlank(username)
```
james-toussaint: isAuthenticated -> shouldAuthenticate?

src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
```diff
@@ -59,10 +81,15 @@ static void purgeClients() {
...
+                && StringUtils.isNotBlank(password);
```
james-toussaint: boolean isAuthenticated can be moved in the if

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -777,18 +811,29 @@ public synchronized boolean removeContainer(String containerName) {
...
+     * @return an authenticated docker client if credentials are provided
+     * @throws IllegalArgumentException if registry address is blank
+     * @throws DockerException if authentication fails
+     */
+    private static DockerClient createClient(String registryAddress, String username,
+            String password) throws DockerException, IllegalArgumentException {
+        if (StringUtils.isBlank(registryAddress)) {
+            throw new IllegalArgumentException("Registry address must not be blank");
+        }
+        boolean authenticate = StringUtils.isNotBlank(username)
```
james-toussaint: Same, shoudAuthenticate maybe?

src/main/java/com/iexec/common/docker/client/DockerClientFactory.java
zguesmi: Actually, it is already checked inside `DockerClientInstance` constructor. It would be redundant here. I'll also remove the check in the other method [here](https://github.com/iExecBlockchainComputing/iexec-common/pull/249/files#diff-68d2a5a4981136e67af7a2df423f7cb358eac08e545c54a57268fdc04f4e70a4R51).

james-toussaint: :+1:

james-toussaint (approved):

PR #256, opened 2021-08-19
Feature/tx watch frequency

james-toussaint:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -104,6 +108,26 @@ private static int scoreToWeight(int workerScore) {
      * the last ContractGasProvider which depends on the gas price of the network
      */
     public IexecHubContract getHubContract(ContractGasProvider contractGasProvider) {
+        return getHubContract(contractGasProvider,
+                ChainIdLong.NONE,
```
zguesmi, line 112: Will this method still be used?

zguesmi (approved):

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
james-toussaint: This is the (updated but) same method than before, without EIP155 protection, used everywhere (worker, legacy core, products, ..) for sending txs.

PR #261, opened 2021-09-21
Bump to next version

zguesmi:


build.gradle
```diff
@@ -76,7 +76,7 @@ ext {
     isMasterBranch = gitBranch == "master"
     isDevelopBranch = gitBranch == "develop"
     canUploadArchives = isMasterBranch || isDevelopBranch
-    isSnapshotVersion = project.version.contains("SNAPSHOT")
+    isSnapshotVersion = project.version.contains("NEXT")
```
mcornaton, line 79: Should we keep `isSnapshotVersion` name? We could probably rename it to `isNextVersion`.

mcornaton (approved):

build.gradle
zguesmi: I found `isSnapshotVersion` more intuitive because if it is the case we push to the snapshots repo of nexus https://github.com/iExecBlockchainComputing/iexec-common/blob/develop/build.gradle#L114

PR #262, opened 2021-09-27
Add `ReplicateStatus#isRunningFailure` method

mcornaton:
Add a "running failure" Replicate meta-status: it includes all failures that can be encountered before a `COMPUTED` status has been reached.

### DIFF
--- a/src/main/java/com/iexec/common/replicate/ReplicateStatus.java
+++ b/src/main/java/com/iexec/common/replicate/ReplicateStatus.java
@@ -67,7 +67,7 @@ public enum ReplicateStatus {
         return getFailureStatuses().contains(status);
     }
 
-    public static boolean isRunningFailure(ReplicateStatus status) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/iexec/common/replicate/ReplicateStatus.java
```diff
@@ -67,6 +67,10 @@ public static boolean isFailure(ReplicateStatus status) {
         return getFailureStatuses().contains(status);
     }
 
+    public static boolean isRunningFailure(ReplicateStatus status) {
```
james-toussaint: isRunningFailure -> isFailedBeforeComputed
(1) running is dedicated to task, plus (2) more explicit I would say

mcornaton: You're right. The method has been renamed :)

zguesmi (approved):

PR #264, opened 2021-10-04
Hotfix/reuse web3j instance

james-toussaint:


src/main/java/com/iexec/common/chain/Web3jAbstractService.java
```diff
@@ -66,7 +66,9 @@ public static BigInteger getMaxTxCost(long gasPriceCap) {
     }
 
     public Web3j getWeb3j(boolean shouldCheckConnection) {
-        web3j = Web3j.build(new HttpService(chainNodeAddress));
+        if (web3j == null){
+            web3j = Web3j.build(new HttpService(chainNodeAddress));
+        }
```
FabLafarge: Warning : 
- not thread safe (two simultaneous calls may still get 2 separate instances of Web3J if still unitialized)
- separate subclasses/instances of Web3AbstractService will still have separate instances of Web3J
What about Web3j being a true module wide singleton ?

james-toussaint: * What do you think about a double-checked locking with volatile? (topic A)
* Yes, but the Web3AbstractService is intended to be instantiated only once per application, so this warning can be discarded IMO (topic B)
What do you think?

FabLafarge: - I dislike double-check locks (ugly, verbose boiler plate code) and usually prefer initializing into the constructor if lazyness is not an issue (in the best case scenario, I'd let dependency injection handle that), but that's an acceptable solution.
- If only one Web3jAbstractService at runtime, then by all means discard remark B

mcornaton (approved):
This seems to ensure uniqueness of Web3J instance. Great!

FabLafarge (approved):

PR #271, opened 2021-11-03
Add app id to TaskDescription

mcornaton:


src/main/java/com/iexec/common/task/TaskDescription.java
```diff
@@ -42,6 +42,7 @@
     private String callback;
     private DappType appType;
     private String appUri;
+    private String appId;
```
james-toussaint: You can use appAddress wording

gradle.properties
```diff
@@ -1,3 +1,3 @@
-version=5.6.0-NEXT-SNAPSHOT
+version=5.6.1-NEXT-SNAPSHOT
```
james-toussaint: Our current standard for develop version is `<freshest master tag>-NEXT-SNAPSHOT`, so you should not bump here

zguesmi (approved):

james-toussaint (approved):
Thanks

PR #272, opened 2021-11-09
Merge abort notifications

zguesmi:


src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -850,4 +869,13 @@ private static DockerClient createClient(String registryAddress, String username
...
     }
+
+    /**
+     * Get the list of statuses where a container is considered active.
+     * 
+     * @return List.of("running", "restarting")
+     */
+    private List<String> getContainerActiveStatuses() {
+        return List.of(RESTARTING_STATUS, RUNNING_STATUS);
+    }
```
mcornaton: Is there any reason to use a function instead of a constant?

src/main/java/com/iexec/common/replicate/ReplicateStatusCause.java
```diff
@@ -56,4 +58,16 @@
...
     ABORTED_BY_WORKER,
     CANNOT_REVEAL,
+    UNKNOWN;
+
+    public static ReplicateStatusCause getReplicateAbortCause(TaskAbortCause cause) {
+        switch (cause) {
+            case CONSENSUS_REACHED:
+                return CONSENSUS_REACHED;
+            case CONTRIBUTION_TIMEOUT:
+                return CONTRIBUTION_TIMEOUT;
```
mcornaton: I find it hard to read as we have `TaskAbortCause` and `ReplicateStatusCause` with the same member names. Would it be better to statically qualify them?

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
zguesmi: At the beginning I thought I'd need to mock it, but know I think it can be removed. Thanks for the reminder.

src/main/java/com/iexec/common/replicate/ReplicateStatusCause.java
zguesmi: Of course :+1:

src/main/java/com/iexec/common/notification/TaskNotificationExtra.java
```diff
@@ -34,4 +35,8 @@
     // block number from which this notification makes sense
     // (it is not used for all notification types)
     long blockNumber;
+
+    // The reason behind an "Abort" notification. Used only for
+    // Abort notifications.
+    TaskAbortCause taskAbortCause;
```
james-toussaint: Why not private?

zguesmi: Just an oversight. Thanks!

mcornaton (approved):
Great, thanks! :)

james-toussaint (approved):

mcornaton (approved):
Once the tests are green, let's merge! :)

PR #274, opened 2021-11-10
Customized block time in IexecHubAbstractService

mcornaton:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -115,6 +113,7 @@ public IexecHubAbstractService(Credentials credentials,
         this.web3jAbstractService = web3jAbstractService;
         this.iexecHubAddress = iexecHubAddress;
         this.nbBlocksToWaitPerRetry = nbBlocksToWaitPerRetry;
+        this.blockTime = blockTime;
```
james-toussaint: `blockTime > 0 ? blockTime: DEFAULT_BLOCK_TIME` (same applies when a DEFAULT_BLOCK_TIME appears)

mcornaton: Good idea, thanks!

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -109,12 +110,37 @@ public IexecHubAbstractService(Credentials credentials,
...
+                                   Duration blockTime,
+                                   int nbBlocksToWaitPerRetry,
+                                   int maxRetries) {
         this.credentials = credentials;
         this.web3jAbstractService = web3jAbstractService;
         this.iexecHubAddress = iexecHubAddress;
         this.nbBlocksToWaitPerRetry = nbBlocksToWaitPerRetry;
-        this.blockTime = blockTime;
-        this.retryDelay = nbBlocksToWaitPerRetry * blockTime;
+        this.blockTime = blockTime.toMillis() > 0 ? blockTime : Duration.ofMillis(DEFAULT_BLOCK_TIME);
```
james-toussaint: What if blocktime is null?

mcornaton: Well... NPE. This should probably not happen, but here is a fix.

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -111,11 +110,43 @@ public IexecHubAbstractService(Credentials credentials,
...
+                                   int nbBlocksToWaitPerRetry,
+                                   int maxRetries) {
         this.credentials = credentials;
         this.web3jAbstractService = web3jAbstractService;
         this.iexecHubAddress = iexecHubAddress;
         this.nbBlocksToWaitPerRetry = nbBlocksToWaitPerRetry;
-        this.retryDelay = nbBlocksToWaitPerRetry * blockTime;
+        if (blockTime == null || blockTime.toMillis() <= 0) {
+            log.warn("Blocktime value is incorrect, using default value [blockTime:{}, DEFAULT_BLOCK_TIME:{}]",
+                    blockTime == null ? "null" : blockTime, DEFAULT_BLOCK_TIME);
```
zguesmi: There is no need for this
```suggestion
                    blockTime, DEFAULT_BLOCK_TIME);
```

mcornaton: You're right! Updated, thanks :)

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -111,11 +110,43 @@ public IexecHubAbstractService(Credentials credentials,
                                    int blockTime,
                                    int nbBlocksToWaitPerRetry,
                                    int maxRetries) {
+        this(
+                credentials,
+                web3jAbstractService,
+                iexecHubAddress,
+                Duration.ofMillis(blockTime > 0 ? blockTime : DEFAULT_BLOCK_TIME),
```
james-toussaint: I don't think we need that. If block time is not > 0, below base constructor will build duration with DEFAULT_BLOCK_TIME.
I would also add a deprecation tag on this method (comment needs to be updated too, it is not the base constructor anymore).

mcornaton: Right, thanks.
I was also thinking about deprecating this constructor but was unsure about that. So thanks again for pointing this out :)

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -97,25 +96,57 @@ public IexecHubAbstractService(Credentials credentials,
...
+                                   Duration blockTime,
+                                   int nbBlocksToWaitPerRetry,
+                                   int maxRetries) {
         this.credentials = credentials;
         this.web3jAbstractService = web3jAbstractService;
         this.iexecHubAddress = iexecHubAddress;
         this.nbBlocksToWaitPerRetry = nbBlocksToWaitPerRetry;
-        this.retryDelay = nbBlocksToWaitPerRetry * blockTime;
+        if (blockTime == null || blockTime.toMillis() <= 0) {
+            log.warn("Blocktime value is incorrect, using default value [blockTime:{}, DEFAULT_BLOCK_TIME:{}]",
```
james-toussaint: Line should wrapped after 80

mcornaton: Right, that's fixed. My IDE settings were wrong :sweat_smile:

james-toussaint (approved):

zguesmi (approved):

PR #276, opened 2021-11-12
Feature/fix skipped should pull private image test

jbern0rd:


src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
```diff
@@ -449,29 +448,23 @@ public void shouldNotPullImageSinceDockerCmdException() {
...
+     * Try to pull a private image from iexechub, require valid login and permissions.
+     * The test will fail if Docker Hub credentials are missing or invalid.
      */
-    // TODO fix this test
     @Test
-    @Disabled
     public void shouldPullPrivateImage() throws Exception {
         String username = getEnvValue(DOCKERHUB_USERNAME_ENV_NAME);
         String password = getEnvValue(DOCKERHUB_PASSWORD_ENV_NAME);
-        if (StringUtils.isEmpty(username) || StringUtils.isEmpty(password)) {
```
mcornaton, line 460: I think it'd be better to keep this test and call `fail("No dockerhub username or password found, aborting shouldPullPrivateImage test")` instead of simply logging the message. It would make debugging easier than relying on `pullImage` behavior.
By the way, that could be another test: check that pulling a private image does not work without right credentials.

jbern0rd: Hmm, I'm not sure about this one. After deleting this check, we let the `docker-java` library perform its own check and raise its own error.
The `shouldPullPrivateImage` implementation becomes more or less the same as in `shouldGetAuthenticatedClientWithDockerIoRegistry`.
If **credentials** are both present and valid, the test passes, else, it fails.
There is no added value to do a check in test code before creating the DockerClientInstance instance.
This is the rational behind my decision to delete the code.

At the end, I don't really know if those tests are at the right place, because we are actually validating our integration of `docker-java`. It seems we are talking about integration tests here, not unit tests.

mcornaton: > At the end, I don't really know if those tests are at the right place, because we are actually validating our integration of docker-java. It seems we are talking about integration tests here, not unit tests.

You're probably right. FYI, at some point we only had unit tests in the projects so it looks like every test not related to multiple projects - common, core, worker - went there. We have started to build integration related to single projects directly into them - blockchain adapter, Oracle Factory, ... - but that's still a work in progress. Maybe we should move the tests you are mentioning into a new `iexec-common` integration tests module.
Let's wait for Jérémy & Zied's opinions on this point.

james-toussaint: (1) Reason for optional execution on CI/CD of this particular test is time and debt. If it runs on CI/CD it is great.
(2) Reason for optional execution on developer machines of this particular test is speed and simplicity. The test class already has a "slow" tag, so won't be executed with -Pdev option when building. so we can remove `if creds empty skip test`.

james-toussaint: About UT or IT concerns, Docker tests are currently in UT for simplicity, but most of them are calling external Docker daemon and should be moved to dedicated IT instead.
It worths mentioning the single iexec-common library might be split in multiple ones in the future (separating concerns).

james-toussaint (approved):
Thanks :+1:

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
jbern0rd: I forgot a precision, by removing the `if creds empty skip test`, the behavior becomes `if creds empty fail test` which is the same behavior as in `shouldGetAuthenticatedClientWithDockerIoRegistry`.
Both test methods should succeed or fail at the same time now.

zguesmi: Thanks for the fix @jbern0rd, well done. The idea behind the check of username & password is to be able to run tests locally when Dockerhub credentials are not present. Maybe we can tag the test as "ci" and ignore it when we run tests locally.

mcornaton (approved):
:+1:

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
jbern0rd: What would you prefer :
1. A tag on the `shouldPullPrivateImage` test method ?
2. A tag for all test methods at the class level ?

Won't that be redundant with the `slow` tag and the `-Pdev` property ?

zguesmi (approved):
Well done!

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
james-toussaint: I tend to say it is redundant with slow tag.

james-toussaint: If I am working on a test class featuring slow tag, I don't want to be lazy, I want to execute them all.

PR #277, opened 2021-11-16
Add data to `PublicChainConfig`

mcornaton:


src/main/java/com/iexec/common/config/PublicChainConfig.java
```diff
@@ -31,5 +31,12 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class PublicChainConfig {
+    private Integer chainId;
+    private boolean isSidechain;
+    private String nodeAddress;
+    private String hubAddress;
     private Duration blockTime;
+    private long startBlockNumber;
```
james-toussaint: I don't think we need 
```
    private long startBlockNumber;
    private float gasPriceMultiplier;
    private long gasPriceCap;
```

mcornaton: Depending on what we want to achieve, either we let these values into core, or we get it from the blockchain adapter. I think it is better to have the blockchain adapter act as a configuration server and gather all the public blockchain data in this class. Don't you agree?

james-toussaint: startBlockNumber is related to dealwatching, we don't plan to migrate it in short time.
gas* are here for people wanting to tune there tx costs. These values are totally up to people, we don't want to force them using remote ones.

mcornaton: OK, I see. They are removed!

james-toussaint:
PublicConfiguration should be shrinked since partial content is moved to the new class

mcornaton:
Sure, it'll be once core PR is ready.

zguesmi (approved):
Thanks :+1:

james-toussaint (approved):
Thanks for the update!

PR #279, opened 2021-11-18
Feature/hex string utils

james-toussaint:


src/main/java/com/iexec/common/utils/BytesUtils.java
```diff
@@ -44,9 +45,32 @@ public static String hexStringToAscii(String hexString) {
...
+     * Hexadecimal string input must be lowercase and must contain 0x prefix.
+     *
+     * @param hexString        hexadecimal input
+     * @param expectedByteSize expected byte size
+     * @return true if
+     */
+    public static boolean isHexStringWithProperByteSize(String hexString,
+                                                        int expectedByteSize) {
+        return expectedByteSize > 0
+                && !StringUtils.isEmpty(hexString)
```
zguesmi:
```suggestion
                && StringUtils.isNotEmpty(hexString)
```

src/main/java/com/iexec/common/utils/BytesUtils.java
```diff
@@ -44,9 +45,32 @@ public static String hexStringToAscii(String hexString) {
...
+                && Numeric.containsHexPrefix(hexString)
+                && isHexaString(hexString)
+                && BytesUtils.stringToBytes(hexString).length == expectedByteSize;
```
zguesmi: I find method `BytesUtils.stringToBytes(hexString)` confusing. The name gives the impression that it has the same behavior as `String.getBytes()` but it is not. I think it's better to either directly use `Numeric.hexStringToByteArray(hexString)` or rename it to something like `hexStringToBytes()`. I tend to prefer the first option as the method itself seems to not be very useful.

src/test/java/com/iexec/common/utils/BytesUtilsTests.java
```diff
@@ -61,6 +61,50 @@ public void shoudNotBeHexaStringSinceEmpty() {
...
+        assertFalse(isHexStringWithProperByteSize("", 6));
+    }
+
+    @Test
+    void shouldNotBeHexStringWithProperByteSizeSinceNoPrefix() {
+        assertFalse(isHexStringWithProperByteSize("abc123", 6));
+    }
+
+    @Test
+    void shouldNotBeHexStringWithProperByteSizeSinceNotLowerCase() {
```
zguesmi: Why is that? In this case shouldn't the method be named `isLowercaseHexStringWithProperByteSize()`

james-toussaint: This is what I was updating, let me send you a next version

src/main/java/com/iexec/common/utils/BytesUtils.java
```diff
@@ -44,9 +46,42 @@ public static String hexStringToAscii(String hexString) {
     }
 
     public static boolean isHexaString(String hexaString) {
-        return Numeric.cleanHexPrefix(hexaString).matches("\\p{XDigit}+"); // \\p{XDigit} matches any hexadecimal character
+        return !StringUtils.isEmpty(hexaString)
```
mcornaton:
```suggestion
        return StringUtils.isNotEmpty(hexaString)
```

mcornaton (approved):

src/main/java/com/iexec/common/utils/BytesUtils.java
```diff
@@ -44,9 +46,43 @@ public static String hexStringToAscii(String hexString) {
...
+                new byte[expectedByteSize]);
+    }
+
+    public static boolean isBytes32(String hexString) {
+        return isHexStringWithProperBytesSize(hexString, 32);
+    }
+
+    public static boolean isNonEmptyBytes32(String hexString) {
+        return isHexStringWithNonEmptyProperBytesSize(hexString, 32);
+    }
```
jbern0rd, line 98: Should we introduce a `private static final int PROPER_BYTES_SIZE = 32` numeric constant ?

src/main/java/com/iexec/common/utils/BytesUtils.java
```diff
@@ -44,9 +46,43 @@ public static String hexStringToAscii(String hexString) {
...
+     *
+     * @param hexString        hexadecimal input
+     * @param expectedByteSize expected byte size
+     * @return true if
+     */
+    public static boolean isHexStringWithProperBytesSize(String hexString,
+                                                        int expectedByteSize) {
+        return expectedByteSize > 0
+                && StringUtils.isNotEmpty(hexString)
+                && Numeric.containsHexPrefix(hexString)
```
jbern0rd: `StringUtils.isNotEmpty` will be called another time in `isHexaString`.

jbern0rd: Why not simply `hexString.matches("^0x\\p{XDigit}+$")` ?
If there is a match:
* it is not empty
* it has the prefix
* we don't need to remove the prefix to check if it is correct

Or am I missing something entirely ?

jbern0rd: For improved performance, the best implementation should look like:

`private static final Pattern hexPattern = Pattern.compile(<regex>);`

then

`hexPattern.matcher(hexString).matches()`

This avoids to compile the regex pattern every time `isHexaString` is called

src/main/java/com/iexec/common/utils/BytesUtils.java
```diff
@@ -44,9 +46,43 @@ public static String hexStringToAscii(String hexString) {
...
+                && isHexaString(hexString)
+                && Numeric.hexStringToByteArray(hexString).length == new byte[expectedByteSize].length;
```
zguesmi: Why not?
```suggestion
                && Numeric.hexStringToByteArray(hexString).length == expectedByteSize;
```

zguesmi (approved):

zguesmi:
Good :+1:

mcornaton (approved):
LGTM :)

src/main/java/com/iexec/common/utils/BytesUtils.java
james-toussaint: 8 bits = 1 byte -> 2 char slots (0xff fits in 1 byte)?

mcornaton: Not sure about what you mean, but `new byte[expectedByteSize].length` is strictly equal to `expectedByteSize` as `new type[n]` creates an array of length `n` and `array.length`retrieves the length of the array...

james-toussaint: "new byte[expectedByteSize].length is strictly equal to expectedByteSize"
totally agree, let me update that :+1:

src/main/java/com/iexec/common/utils/BytesUtils.java
james-toussaint: I agree, updated thanks

src/main/java/com/iexec/common/utils/BytesUtils.java
james-toussaint: Updated thanks

src/main/java/com/iexec/common/utils/BytesUtils.java
james-toussaint: I agree. Updated, how about that https://github.com/iExecBlockchainComputing/iexec-common/pull/279/commits/615c56c352054102b665c6a8575900310edf44ab?

jbern0rd: :+1: 
Looks great to me, thx !

PR #280, opened 2021-11-18
Feature/update publishing plugin

jbern0rd:
Use 'maven-publish' plugin.
Use `./gradlew clean publishToMavenLocal`, then retrieve dependency in iexec-core or iexec-worker without editing `build.gradle.`

Project moved to `java-library` plugin to write POM with web3j dependencies at `compile` maven scope.
This is done with the new `api` configuration.

It is kind of a dirty fix.
The right way would be to create a platform project and to manage a BOM like it is done in Spring framework.
But this is entirely another beast to handle for the moment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,10 +20,6 @@ repositories {
     }
 }
 
-configurations {
-    deployerJars
-}
-
 dependencies {
     // web3j dependency
     implementation "org.web3j:core:4.7.0"
@@ -67,9 +63,6 @@ dependencies {
     testImplementation 'org.mockito:mockito-core:3.6.0'
     testImplementation 'org.mockito:mockito-junit-jupiter:3.6.0'
     testImplementation 'org.assertj:assertj-core:3.18.1'
-
-    // package cloud
-    deployerJars "io.packagecloud.maven.wagon:maven-packagecloud-wagon:0.0.6"
 }
 
 ext {
--- a/build.gradle
+++ b/build.gradle
@@ -105,9 +105,8 @@ jar {
 }
 
 shadowJar { //required if we want to embed web3j dependency inside
-    baseName = rootProject.name
+    archiveClassifier.set('all')
     from sourceSets.main.allSource
-    version = version
 }
 build.dependsOn(shadowJar)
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -20,28 +20,27 @@ repositories {
...
-    deployerJars
-}
-
+// java-library plugin defines 'api' configuration
+// 'api' configuration allows to expose dependencies with 'compile' scope in pom
+// 'implementation' configuration allows to expose dependencies with 'runtime' scope in pom
 dependencies {
     // web3j dependency
-    compile "org.web3j:core:4.7.0"
+    api "org.web3j:core:4.7.0"
```
james-toussaint, line 28: Nice!

build.gradle
```diff
@@ -20,28 +20,27 @@ repositories {
...
-    compile 'javax.xml.bind:jaxb-api:2.3.0'
+    implementation 'javax.xml.bind:jaxb-api:2.3.0'
 
     //args utils
-    compile 'org.apache.maven.shared:maven-shared-utils:3.2.1'
+    implementation 'org.apache.maven.shared:maven-shared-utils:3.2.1'
 
     // zip
-    compile 'net.lingala.zip4j:zip4j:2.3.1'
+    implementation 'net.lingala.zip4j:zip4j:2.3.1'
```
james-toussaint, line 43: Do you confirm we don't need it in worker?

james-toussaint:
Don't you think we only need a BOM at the moment we split iexec-common into multiple sub-libraries?

build.gradle
jbern0rd: `grep -r 'zip4' src` is empty in iexec-worker.
I only see calls to com.iexec.common.utils.FileHelper and com.iexec.common.worker.result.ResultUtils.
We should be clean.

jbern0rd:
We could already benefit from a BOM, but it's not that urgent.
I think the same as you, it will have even more value if there is a split into multiple sub-libraries.

build.gradle
james-toussaint: Ok :+1:

james-toussaint (approved):

PR #281, opened 2021-11-19
Expose public chain configuration

mcornaton:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -149,13 +171,47 @@ public IexecHubAbstractService(Credentials credentials,
...
+        this.expectedFinalDeadlineRatio = expectedFinalDeadlineRatio;
 
         String hubAddress = getHubContract().getContractAddress();
         log.info("Abstract IexecHubService initialized (iexec proxy address) " +
                 "[hubAddress:{}]", hubAddress);
         setMaxNbOfPeriodsForConsensus();
     }
 
+    @PostConstruct
+    private void validateChainConfiguration() {
```
james-toussaint: I would split the validation in 2 parts:
1. A node exists and can reply (see rpc call to getNodeVersion)
2. A business check like you do, shortly verifying we not connected to a wrong smartcontract (what you already do)

About (2.), I'm not very van of the semantic but I don't have good suggestions so far. The question is "Why introducing this particular field?".

james-toussaint: Maybe we could wrap `expectedFinalDeadlineRatio` into something like a `IexecSmartContractConnectionPolicy` class?

james-toussaint: Part (1.) might be done in `Web3jService`

src/main/java/com/iexec/common/config/PublicChainConfig.java
```diff
@@ -31,5 +31,9 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class PublicChainConfig {
+    private Integer chainId;
+    private boolean isSidechain;
+    private String nodeAddress;
+    private String hubAddress;
```
zguesmi: I find it clearer this way
```suggestion
    private String chainNodeUrl;
    private String hubContractAddress;
```

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
mcornaton: That looks way better, thanks. New version is online, let me know if it suits you :)

BTW, I haven't seen your last comment before uploading the new version. I'll look into it this afternoon.

src/main/java/com/iexec/common/config/PublicChainConfig.java
james-toussaint: I agree, how about `iExecHubContractAddress`?

mcornaton: `iExecHubContractAddress` would be better as we specifically target our contract. Thanks for the suggestions!

src/main/java/com/iexec/common/contract/IexecSmartContractConnectionPolicy.java
```diff
@@ -0,0 +1,62 @@
...
+
+package com.iexec.common.contract;
+
+import com.iexec.common.contract.generated.IexecHubContract;
+import lombok.extern.slf4j.Slf4j;
+
+import java.math.BigInteger;
+
+@Slf4j
+public class IexecSmartContractConnectionPolicy {
```
zguesmi: Isn't it more of a `IexecHubSmartContractValidator/Checker`?
Also `expectedFinalDeadlineRatio` is a constant (== 10) that depends on the contract, so I think it can be encapsulated in the class and there is no need to pass it around as an argument.

mcornaton: `Validator` would be great as it would look like [`TeeEnclaveConfigurationValidator` class](https://github.com/iExecBlockchainComputing/iexec-common/blob/develop/src/main/java/com/iexec/common/tee/TeeEnclaveConfigurationValidator.java).

I'd like to encapsulate `expectedFinalDeadlineRatio` in the class, but it looks like there's currently no value injection in `iexec-common`. I was unsure whether it was intentional so I didn't want to bother with it. Do you think it'd be great to add this feature to Common?

zguesmi: I mean adding a constant `EXPECTED_FINAL_DEADLINE_RATIO` in the class `IexecSmartContractValidator`.

mcornaton: @jeremyjams suggested that this value should be parametrized so that we could update it if evolves one day. I agree passing it through the whole chain is not really clean but I haven't find any better solution :(

james-toussaint: 1. `IexecHubSmartContractValidator` is a good name.
2. In addition to the existing one, I suggest you to add a default empty constructor which uses an already set `EXPECTED_FINAL_DEADLINE_RATIO`.

src/main/java/com/iexec/common/contract/IexecSmartContractConnectionPolicy.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        if (expectedFinalDeadlineRatio == null) {
+            log.warn("Can't check final deadline ratio " +
+                    "as no expected value is provided." +
+                    "[actualValue: {}]", finalDeadlineRatio);
+            return;
+        }
+
+        if (!finalDeadlineRatio.equals(BigInteger.valueOf(expectedFinalDeadlineRatio))) {
```
james-toussaint: NPE?

src/main/java/com/iexec/common/contract/IexecSmartContractConnectionPolicy.java
```diff
@@ -0,0 +1,62 @@
...
+
+@Slf4j
+public class IexecSmartContractConnectionPolicy {
+    private final Integer expectedFinalDeadlineRatio;
+
+    public IexecSmartContractConnectionPolicy(Integer expectedFinalDeadlineRatio) {
+        this.expectedFinalDeadlineRatio = expectedFinalDeadlineRatio;
+    }
+
+    public void checkSmartContractConnection(IexecHubContract contract) {
```
james-toussaint:
```suggestion
    public boolean validate(IexecHubContract contract) {
```

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -149,13 +172,22 @@ public IexecHubAbstractService(Credentials credentials,
...
 
         String hubAddress = getHubContract().getContractAddress();
         log.info("Abstract IexecHubService initialized (iexec proxy address) " +
                 "[hubAddress:{}]", hubAddress);
         setMaxNbOfPeriodsForConsensus();
     }
 
+    @PostConstruct
+    private void checkSmartContractConnection() {
+        iexecSmartContractConnectionPolicy.checkSmartContractConnection(
```
james-toussaint: throw if validate == false

src/main/java/com/iexec/common/contract/IexecSmartContractConnectionPolicy.java
zguesmi: @jeremyjams If the value evolves, that means the smart contract also has evolved and we need to update iexec-common in all cases. So I find it coherent to use the constant.

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -149,13 +171,22 @@ public IexecHubAbstractService(Credentials credentials,
...
 
         String hubAddress = getHubContract().getContractAddress();
         log.info("Abstract IexecHubService initialized (iexec proxy address) " +
                 "[hubAddress:{}]", hubAddress);
         setMaxNbOfPeriodsForConsensus();
     }
 
+    @PostConstruct
+    private void checkSmartContractConnection() {
+        iexecSmartContractValidator.checkSmartContractConnection(
```
zguesmi: Careful, in one the constructors (the [deprecated](https://github.com/iExecBlockchainComputing/iexec-common/pull/281/files#diff-fe771d01d22a765a39092fff6798eb71adb079c50f9e0ab2c1088eeafaa39279R138) one) `iexecSmartContractValidator` is null.

src/main/java/com/iexec/common/contract/IexecSmartContractConnectionPolicy.java
mcornaton: Not sure whether `contract.final_deadline_ratio().send()` could return a null value, but let's add this security!

src/main/java/com/iexec/common/contract/IexecSmartContractValidator.java
```diff
@@ -0,0 +1,62 @@
...
+
+package com.iexec.common.contract;
+
+import com.iexec.common.contract.generated.IexecHubContract;
+import lombok.extern.slf4j.Slf4j;
+
+import java.math.BigInteger;
+
+@Slf4j
+public class IexecSmartContractValidator {
```
zguesmi: Iexec**Hub**SmartContractValidator

src/main/java/com/iexec/common/contract/IexecSmartContractConnectionPolicy.java
zguesmi: The other option would be to not have an `expectedValue` and use the value fetched from the smart contract. We only need to make sure it is and int `> 0`.

src/main/java/com/iexec/common/contract/IexecSmartContractValidator.java
mcornaton: Right!

src/main/java/com/iexec/common/contract/IexecSmartContractConnectionPolicy.java
james-toussaint: I agree, what truly matters is `final_deadline_ratio()` call returns something greater than zero (meaning the remote function exists).

mcornaton: I think I agree with @zguesmi. That probably means we don't need those value injections... And that simplifies the whole logic of the thing.

mcornaton: Alright, so let's remove this whole `expectedValue` thing! :)

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
mcornaton: Right, that was one point depending on the `expectedFinalDeadlineRatio` injection. Now it has been removed, we can have a default `iexecHubSmartContractValidator` for all constructors! :)

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -156,6 +158,16 @@ public IexecHubAbstractService(Credentials credentials,
         setMaxNbOfPeriodsForConsensus();
     }
 
+    @PostConstruct
+    private void checkSmartContractConnection() {
```
james-toussaint:
```suggestion
    private void validateRemoteIexecHubSmartContract() {
```

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -156,6 +158,16 @@ public IexecHubAbstractService(Credentials credentials,
...
+        if (!new IexecHubSmartContractValidator().validate(getHubContract())) {
+            throw new IllegalArgumentException(
+                    "Can't validate iExec Smart Contract."
```
james-toussaint:
```suggestion
                    "IexecHub smart contract validation failed."
```

src/main/java/com/iexec/common/contract/IexecHubSmartContractValidator.java
```diff
@@ -0,0 +1,54 @@
...
+        final BigInteger finalDeadlineRatio;
+        final String errorMessage =
+                "Something went wrong with the chain configuration. "
+                        + "Please check your configuration.";
+        try {
+            finalDeadlineRatio = contract
+                    .final_deadline_ratio()
+                    .send();
+        } catch (Exception e) {
+            log.error(errorMessage);
```
james-toussaint:
```suggestion
            log.error(errorMessage, e);
```

src/main/java/com/iexec/common/contract/IexecHubSmartContractValidator.java
```diff
@@ -0,0 +1,54 @@
...
+            return false;
+        }
+
+        if (finalDeadlineRatio.compareTo(BigInteger.ZERO) <= 0) {
+            log.error(errorMessage + " Final deadline ratio should be positive"
+                    + " [finalDeadlineRatio: {}]", finalDeadlineRatio);
+            return false;
+        }
+
+        log.info("Connection to iExec smart contract has been established.");
```
james-toussaint:
```suggestion
        log.info("IexecHub smart contract is reachable.");
```

src/main/java/com/iexec/common/contract/IexecHubSmartContractValidator.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.math.BigInteger;
+
+@Slf4j
+public class IexecHubSmartContractValidator {
+    public boolean validate(IexecHubContract contract) {
+        final BigInteger finalDeadlineRatio;
+        final String errorMessage =
+                "Something went wrong with the chain configuration. "
```
james-toussaint:
```suggestion
                "Something went wrong with IexecHub smart contract configuration. "
```

src/main/java/com/iexec/common/config/PublicChainConfig.java
```diff
@@ -31,5 +31,9 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class PublicChainConfig {
+    private Integer chainId;
+    private boolean isSidechain;
+    private String chainNodeUrl;
+    private String iExecHubContractAddress;
```
james-toussaint:
```suggestion
    private String iexecHubContractAddress;
```
lowercase(IexecHub)

james-toussaint (approved):
Very clean :muscle:

zguesmi (approved):

PR #282, opened 2021-11-24
Get address from private key with valid length

james-toussaint:


src/main/java/com/iexec/common/utils/CredentialsUtils.java
```diff
@@ -27,8 +25,11 @@ private CredentialsUtils() {
     }
 
     public static String getAddress(String privateKey) {
-        ECKeyPair ecKeyPair = ECKeyPair.create(BytesUtils.stringToBytes32(privateKey));
-        return Numeric.prependHexPrefix(Keys.getAddress(ecKeyPair));
+        if (BytesUtils.isNonZeroedHexStringWithPrefixAndProperBytesSize(privateKey,
+                BytesUtils.BYTES_32_SIZE)) {
+            return Credentials.create(privateKey).getAddress();
+        }
+        return "";
```
mcornaton, line 37: I think it would be better to log something if we return an empty string. This will probably make the next operations fail at some point, so it'd be great to know where it comes from.

james-toussaint: Updated :+1:

mcornaton (approved):
Thanks! :)

zguesmi (approved):

PR #289, opened 2021-12-02
Add common `ApiResponse`

mcornaton:


james-toussaint (approved):
Let's wait for `ResponseEntity<ApiResponse<...>>` :eyes:  !

src/main/java/com/iexec/common/web/ApiResponse.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+
+import java.util.List;
+
+@Getter
+@Builder
+@EqualsAndHashCode
+@ToString
```
zguesmi: [@Data](https://projectlombok.org/features/Data)

src/main/java/com/iexec/common/web/ApiResponse.java
```diff
@@ -0,0 +1,33 @@
...
+public class ApiResponse<T> {
```
zguesmi:
```suggestion
public class ApiResponseBody<T> {
```
?
Since it's more of a body than a hole response.

zguesmi (approved):

src/main/java/com/iexec/common/web/ApiResponse.java
mcornaton: No setter will be generated as fields are final so that's what we're looking for. Thanks! :)

src/main/java/com/iexec/common/web/ApiResponse.java
mcornaton: Sure, that's more explicit. Thanks!

PR #292, opened 2021-12-10
Feature/build javadoc sources jar

jbern0rd:
Use Gradle mechanisms to produce javadoc and sources artifacts.
Remove copy of source in library jar.
When referencing the iexec-common dependency, this allows to navigate through source code like the other mainstream libraries.
Fixed the javadoc errors, hyperlinks need to be wrapper in anchor tags.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,11 @@ ext {
     isDev = project.hasProperty('dev')
 }
 
+java {
+    withJavadocJar()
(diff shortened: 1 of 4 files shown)


zguesmi (approved):
:+1:

mcornaton (approved):
:+1:

james-toussaint (approved):
Thanks for doc browsing!

PR #293, opened 2021-12-15
Get EIP712Domain without verifying contract if not provided

james-toussaint:


mcornaton (approved):
LGTM, thanks :)

zguesmi (commented):
Thanks for the quick fix.
We need to update it [here](https://github.com/iExecBlockchainComputing/iexec-common/blob/develop/src/main/java/com/iexec/common/chain/eip712/EIP712Domain.java#L44) also.

zguesmi:
I suggest we also add `@Deprecated` to this class so we have an explicit warning.

zguesmi (approved):
Perfect, thansk!

PR #295, opened 2021-12-23
Add `ContextualLockRunner`

mcornaton:
Add an object capable of locking some operation based on a key.
Useful for synchronization issue where we want to select which operations can run in parallel while other should wait their turn.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,10 +21,10 @@ ext {
 
 repositories {
     mavenCentral()
-    jcenter()
     maven {
         url "https://jitpack.io"
     }
+    jcenter()
 }
 
 // java-library plugin defines 'api' configuration
@@ -75,9 +75,7 @@ dependencies {
     implementation 'net.jodah:expiringmap:0.5.10'
 
     // test
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
-    testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.6.0'
-    testImplementation 'org.mockito:mockito-core:3.6.0'
+    testImplementation 'org.junit.jupiter:junit-jupiter:5.6.0'
     testImplementation 'org.mockito:mockito-junit-jupiter:3.6.0'
     testImplementation 'org.assertj:assertj-core:3.18.1'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -60,6 +60,9 @@ dependencies {
     // failsafe
     implementation 'net.jodah:failsafe:2.4.4'
 
+    // expiring map
+    implementation 'net.jodah:expiringmap:0.5.8'
+
     // test
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
     testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.6.0'
--- a/build.gradle
+++ b/build.gradle
@@ -72,7 +72,7 @@ dependencies {
     implementation 'org.awaitility:awaitility:4.1.1'
 
     // expiring map
-    implementation 'net.jodah:expiringmap:0.5.8'
+    implementation 'net.jodah:expiringmap:0.5.10'
 
     // test
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,13 @@ group = 'com.iexec.common'
 sourceCompatibility = 11
 targetCompatibility = 11
 
+ext {
+    // ./gradlew build -Pdev
+    isDev = project.hasProperty('dev')
+    // versions
+    dockerJavaVersion = '3.2.12'
+}
+
 repositories {
     mavenCentral()
     jcenter()
@@ -54,11 +61,15 @@ dependencies {
     implementation 'io.github.openfeign:feign-slf4j:11.0'
 
     // docker
-    implementation 'com.github.docker-java:docker-java:3.2.5'
-    implementation 'com.github.docker-java:docker-java-transport-httpclient5:3.2.5'
+    implementation "com.github.docker-java:docker-java:${dockerJavaVersion}"
+    // Using zerodep until docker client stops blocking after having called
+    // listContainersCmd() over httpclient5
+    implementation "com.github.docker-java:docker-java-transport-zerodep:${dockerJavaVersion}"
 
     // failsafe
     implementation 'net.jodah:failsafe:2.4.4'
+    //TODO Migrate from Awaitility to Failsafe everywhere
+    implementation 'org.awaitility:awaitility:4.1.1'
 
     // expiring map
     implementation 'net.jodah:expiringmap:0.5.8'
@@ -71,10 +82,6 @@ dependencies {
     testImplementation 'org.assertj:assertj-core:3.18.1'
 }
 
-ext {
-    // ./gradlew build -Pdev
-    isDev = project.hasProperty('dev')
-}
 
 java {
     withJavadocJar()
--- a/src/main/java/com/iexec/common/utils/ContextualLockRunner.java
+++ b/src/main/java/com/iexec/common/utils/ContextualLockRunner.java
@@ -30,18 +30,18 @@ import java.util.function.Supplier;
  * but two actions with different keys can be run at the same time.
  * <br>
  * To avoid memory-leaks, entries expire after a given time.
- * Default is 1 hour, but it can be set at {@link TargetedLockRunner} object creation.
+ * Default is 1 hour, but it can be set at {@link ContextualLockRunner} object creation.
  *
  * @param <K> Type of the key.
  */
-public class TargetedLockRunner<K> {
+public class ContextualLockRunner<K> {
     private final ConcurrentMap<K, Object> locks;
 
-    public TargetedLockRunner() {
+    public ContextualLockRunner() {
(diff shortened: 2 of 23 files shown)


src/main/java/com/iexec/common/utils/TargetedLock.java
```diff
@@ -0,0 +1,77 @@
...
+        this(1, TimeUnit.HOURS);
+    }
+
+    public TargetedLock(long expirationDuration, TimeUnit unit) {
+        this.locks = ExpiringMap.builder()
+                .expiration(expirationDuration, unit)
+                .build();
+    }
+
+    public void runWithLock(K key, Runnable action) {
```
james-toussaint:
```
    public void runWithLock(K key, Runnable action) {
        // `ConcurrentHashMap::computeIfAbsent` is atomic, so there can't be any race condition there.
        synchronized (getLock(key)) {
            action.run();
        }
    }
    
    private Object getLock(K key) {
        return locks.computeIfAbsent(key, id -> new Object());
    }
```
?

mcornaton: Good idea, thanks! Added in https://github.com/iExecBlockchainComputing/iexec-common/pull/295/commits/60f5d53197d0ac1cd3f67c29646457e6e7979e6f.

src/main/java/com/iexec/common/utils/TargetedLock.java
```diff
@@ -0,0 +1,80 @@
...
+ * Provides a way to avoid parallel runs of an action.
+ * Locks are based on a key, so that two actions with the same key can't be run at the same time
+ * but two actions with different keys can be run at the same time.
+ * <br>
+ * To avoid memory-leaks, entries expire after a given time.
+ * Default is 1 hour, but it can be set at {@link TargetedLock} object creation.
+ *
+ * @param <K> Type of the key.
+ */
+public class TargetedLock<K> {
```
james-toussaint:
```suggestion
public class LockRunner<K> {
```
?

mcornaton: I'd prefer to keep a reference to `target` as we don't lock everything, just what's related to a target/key.

james-toussaint: `TargetedLockRunner<K>`
?
(A lock is usually binary/passive, but this class does more since it performs some actions)

mcornaton: I agree with you. `TargetedLockRunner` seems better, let's continue with that! Thanks :)
See https://github.com/iExecBlockchainComputing/iexec-common/pull/295/commits/6f6b28ed34dbe12a2186517bd291a9e0bca4034b.

build.gradle
```diff
@@ -71,6 +71,9 @@ dependencies {
     //TODO Migrate from Awaitility to Failsafe everywhere
     implementation 'org.awaitility:awaitility:4.1.1'
 
+    // expiring map
+    implementation 'net.jodah:expiringmap:0.5.8'
```
jbern0rd: Is there a reason to use this version and not the latest one ?

src/test/java/com/iexec/common/utils/TargetedLockRunnerTest.java
```diff
@@ -0,0 +1,89 @@
...
+        final Consumer<Integer> runActionAndGetOrder = (i) -> {
+            for (int j = 0; j < numberOfRunsPerGroup; j++) {
+                actionsOrder.add(i);
+            }
+        };
+
+        // Run the action a bunch of times with different inputs and potentially different keys each time.
+        IntStream.range(0, numberOfActions)
+                .parallel()
+                .forEach(i -> locks.runWithLock(keyProvider.apply(i), () -> runActionAndGetOrder.accept(i)));
```
jbern0rd: Can we check the other methods ? acceptWithLock, getWithLock, applyWithLock, ...

build.gradle
mcornaton: No reason except the fact it's the version we already use in the Result Proxy:
https://github.com/iExecBlockchainComputing/iexec-result-proxy/blob/064b5036afb34397cbd00ce0472158c66c8503ef/build.gradle#L76

But I can certainly use the last version.

src/test/java/com/iexec/common/utils/TargetedLockRunnerTest.java
mcornaton: We could probably test them, yes. But the test algorithm is already really tricky, I'm not sure I want to add tests for such similar methods. Let me think about this.

src/test/java/com/iexec/common/utils/ContextualLockRunnerTest.java
```diff
@@ -0,0 +1,377 @@
...
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.stream.IntStream;
+
+class ContextualLockRunnerTest {
```
zguesmi:
```suggestion
class ContextualLockRunnerTests {
```
Just for consistency's sake.

src/main/java/com/iexec/common/utils/ContextualLockRunner.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    public <R> R getWithLock(K key, Supplier<R> action) {
+        synchronized (getLock(key)) {
+            return action.get();
+        }
+    }
+
+    public <R> R applyWithLock(K key, Function<K, R> action) {
```
zguesmi, line 68: It's a little confusing that `key` is both used to lock and as a parameter of the action. We can maybe add a third argument to the method (`..., T actionArgument`) but I'm afraid it will make the method too complicated to read.

src/test/java/com/iexec/common/utils/ContextualLockRunnerTest.java
mcornaton: Sure, thanks!

src/main/java/com/iexec/common/utils/ContextualLockRunner.java
mcornaton: IMHO, it's better as is. You'll generally want to execute an action on a given key, e.g. when you'll want to run some action and lock it based on a `chainTaskId`, it'll probably be the argument passed to the action.

zguesmi (approved):
Seems good to me. We'll continue iterating on it in the future.
Thanks!

src/main/java/com/iexec/common/utils/ContextualLockRunner.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * Provides a way to avoid parallel runs of an action.
+ * Locks are based on a key, so that two actions with the same key can't be run at the same time
+ * but two actions with different keys can be run at the same time.
+ * <br>
+ * To avoid memory-leaks, entries expire after a given time.
+ * Default is 1 hour, but it can be set at {@link ContextualLockRunner} object creation.
```
jbern0rd: This self-link is a bit redundant.
If the goal is to emphasize, it may be better to use `{@code ContextualLockRunner}` or bold/italic.

src/test/java/com/iexec/common/utils/ContextualLockRunnerTests.java
```diff
@@ -0,0 +1,377 @@
...
+        // As any order is acceptable, there should not be any exception there.
+        Assertions.assertThatCode(() -> runWithoutLock(Function.identity()))
+                .doesNotThrowAnyException();
+    }
+
+    /**
+     * Parity is used to lock all operations:
+     * final order should be a list within which each value should not interfere with others of same parity.
+     * E.g. [2, 1, 2, 2, 0, 1, 1, 3, 0, 0, 0, 3, 3] would be an expected result,
+     * however [2, 1, 2, 2, 1, <b>3</b>, 1, 0, 0, 0, 3, 3] would not be an expected result as 3 interferes with 1.
```
jbern0rd: Are those comments still valid after tests refactoring ?
New algorithm does not seem to check execution order, it seems to check concurrent decrement on a key in a map.
Do not hesitate to correct me if I am wrong

src/main/java/com/iexec/common/utils/ContextualLockRunner.java
mcornaton: That's redundant, you're right! Let's remove it.
https://github.com/iExecBlockchainComputing/iexec-common/pull/295/commits/bbe8cf0050604c871f223ce7ec2c6c589f97093c

src/test/java/com/iexec/common/utils/ContextualLockRunnerTests.java
mcornaton: They are not valid anymore, true. Let's find a new way to describe what's done there.

mcornaton: See https://github.com/iExecBlockchainComputing/iexec-common/pull/295/commits/2c408a0b41f068509502f1eaa50e6257a792359d.

jbern0rd (approved):

PR #296, opened 2021-12-28
Bugfix/docker client with multiple devices

jbern0rd:
Fix implementation where all devices were not correctly added when a container was started with `DockerClientFactory` and `DockerClientInstance`.
See https://github.com/docker-java/docker-java/blob/3.2.5/docker-java-api/src/main/java/com/github/dockerjava/api/model/HostConfig.java#L709

### DIFF
--- a/src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
+++ b/src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
@@ -98,15 +98,16 @@ public class DockerClientInstance {
             throw new IllegalArgumentException("Docker registry password must not be blank");
         }
         this.client = createClient(registryAddress, username, password);
+        log.info("Created authenticated {} docker client {} for {}", registryAddress, client, this);
     }
 
     public DockerClient getClient() {
+        log.info("Use docker client {} for {}", client, this);
         return this.client;
     }
 
-    /**
+    /*
      * Docker volume
-     * 
      */
 
     public synchronized boolean createVolume(String volumeName) {
@@ -175,9 +176,8 @@ public class DockerClientInstance {
         }
     }
 
-    /**
+    /*
      * Docker network
-     * 
      */
 
     public synchronized String createNetwork(String networkName) {
@@ -251,9 +251,8 @@ public class DockerClientInstance {
         }
     }
 
-    /**
+    /*
      * Docker image
-     * 
      */
 
     /**
@@ -359,9 +358,8 @@ public class DockerClientInstance {
         }
     }
 
-    /**
+    /*
      * Docker container
-     * 
      */
 
     /**
@@ -817,9 +815,8 @@ public class DockerClientInstance {
                 .build());
     }
 
-    /**
+    /*
      * Docker client
-     * 
      */
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -563,7 +558,7 @@ public HostConfig buildHostConfigFromRunRequest(DockerRunRequest dockerRunReques
                     dockerRunRequest.getBinds().toArray(new String[0])));
         }
         if (dockerRunRequest.getDevices() != null) {
```
mcornaton: Maybe add a variable to avoid calling twice `dockerRunRequest.getDevices()`?

jbern0rd: See https://github.com/iExecBlockchainComputing/iexec-common/pull/296/commits/5b50661cb8f1122e8edac64d1d06805a1c6c5941

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -557,8 +554,11 @@ public HostConfig buildHostConfigFromRunRequest(DockerRunRequest dockerRunReques
             hostConfig.withBinds(Binds.fromPrimitive(
                     dockerRunRequest.getBinds().toArray(new String[0])));
         }
-        if (dockerRunRequest.getDevices() != null) {
-            hostConfig.withDevices(dockerRunRequest.getDevices());
+        try {
+            List<Device> devices = new ArrayList<>(dockerRunRequest.getDevices());
+            hostConfig.withDevices(devices);
+        } catch (NullPointerException e) {
+            log.info("No devices defined in docker run request.");
```
mcornaton: Why not providing a default empty list instead of risking some NPE later?

jbern0rd: This is the only way to perform only one call to `dockerRunRequest.getDevices()`.
All other implementations will require a check on `dockerRunRequest.getDevices() != null` then `addAll(dockerRunRequest.getDevices())`.
If the goal is to avoid multiple calls to `getDevices`, we *have to* access it once and handle the `NullPointerException` if it happens.

jbern0rd: Implementation without NullPointerException in https://github.com/iExecBlockchainComputing/iexec-common/pull/296/commits/e05ddc496cd2f06a4a5cdd9cd3ec8422b575dea8

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -557,9 +552,8 @@ public HostConfig buildHostConfigFromRunRequest(DockerRunRequest dockerRunReques
             hostConfig.withBinds(Binds.fromPrimitive(
                     dockerRunRequest.getBinds().toArray(new String[0])));
         }
-        if (dockerRunRequest.getDevices() != null) {
-            hostConfig.withDevices(dockerRunRequest.getDevices());
-        }
+        List<Device> devices = dockerRunRequest.getDevices();
```
mcornaton: Nice! What about applying the same idea to both `dockerRunRequest.getXxx()` above?

mcornaton (approved):
:+1:

PR #298, opened 2022-01-03
Feature/provide requester secret refs in deal params

jbern0rd:
Ground work needed to handle requester secret references through `TaskDescription` in `iexec-sms`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,10 +21,10 @@ ext {
 
 repositories {
     mavenCentral()
-    jcenter()
     maven {
         url "https://jitpack.io"
     }
+    jcenter()
 }
 
 // java-library plugin defines 'api' configuration
--- a/build.gradle
+++ b/build.gradle
@@ -61,9 +61,7 @@ dependencies {
     implementation 'net.jodah:failsafe:2.4.4'
 
     // test
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
-    testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.6.0'
-    testImplementation 'org.mockito:mockito-core:3.6.0'
+    testImplementation 'org.junit.jupiter:junit-jupiter:5.6.0'
     testImplementation 'org.mockito:mockito-junit-jupiter:3.6.0'
     testImplementation 'org.assertj:assertj-core:3.18.1'
 }
--- a/src/main/java/com/iexec/common/chain/ChainDeal.java
+++ b/src/main/java/com/iexec/common/chain/ChainDeal.java
@@ -24,7 +24,8 @@ import org.web3j.tuples.generated.Tuple9;
 
 import java.io.IOException;
 import java.math.BigInteger;
-import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 
 @Data
 @NoArgsConstructor
@@ -72,15 +73,19 @@ public class ChainDeal {
         try {
             DealParams dealParams = new ObjectMapper().readValue(params, DealParams.class);
             if (dealParams.getIexecInputFiles() == null) {
-                dealParams.setIexecInputFiles(new ArrayList<>());
+                dealParams.setIexecInputFiles(List.of());
+            }
+            if (dealParams.getIexecSecrets() == null) {
+                dealParams.setIexecSecrets(Map.of());
             }
             return dealParams;
         } catch (IOException e) {
             //the requester want to execute one task with the whole string
             return DealParams.builder()
                     .iexecArgs(params)
-                    .iexecInputFiles(new ArrayList<>())
+                    .iexecInputFiles(List.of())
                     .iexecDeveloperLoggerEnabled(false)
+                    .iexecSecrets(Map.of())
                     .build();
         }
     }
--- a/src/main/java/com/iexec/common/chain/ChainDeal.java
+++ b/src/main/java/com/iexec/common/chain/ChainDeal.java
@@ -24,8 +24,7 @@ import org.web3j.tuples.generated.Tuple9;
 
 import java.io.IOException;
 import java.math.BigInteger;
-import java.util.List;
-import java.util.Map;
+import java.util.Collections;
 
 @Data
 @NoArgsConstructor
@@ -73,19 +72,19 @@ public class ChainDeal {
         try {
             DealParams dealParams = new ObjectMapper().readValue(params, DealParams.class);
             if (dealParams.getIexecInputFiles() == null) {
-                dealParams.setIexecInputFiles(List.of());
+                dealParams.setIexecInputFiles(Collections.emptyList());
             }
             if (dealParams.getIexecSecrets() == null) {
-                dealParams.setIexecSecrets(Map.of());
+                dealParams.setIexecSecrets(Collections.emptyMap());
             }
             return dealParams;
         } catch (IOException e) {
             //the requester want to execute one task with the whole string
             return DealParams.builder()
                     .iexecArgs(params)
-                    .iexecInputFiles(List.of())
+                    .iexecInputFiles(Collections.emptyList())
                     .iexecDeveloperLoggerEnabled(false)
-                    .iexecSecrets(Map.of())
+                    .iexecSecrets(Collections.emptyMap())
                     .build();
         }
     }
--- a/src/main/java/com/iexec/common/chain/ChainDeal.java
+++ b/src/main/java/com/iexec/common/chain/ChainDeal.java
@@ -16,15 +16,12 @@
 
 package com.iexec.common.chain;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.iexec.common.utils.BytesUtils;
 import lombok.*;
 import org.web3j.tuples.generated.Tuple6;
 import org.web3j.tuples.generated.Tuple9;
 
-import java.io.IOException;
 import java.math.BigInteger;
-import java.util.Collections;
 
 @Data
 @NoArgsConstructor
@@ -68,27 +65,6 @@ public class ChainDeal {
                 !getChainDataset().getChainDatasetId().equals(BytesUtils.EMPTY_ADDRESS);
     }
 
-    public static DealParams stringToDealParams(String params) {
-        try {
-            DealParams dealParams = new ObjectMapper().readValue(params, DealParams.class);
-            if (dealParams.getIexecInputFiles() == null) {
-                dealParams.setIexecInputFiles(Collections.emptyList());
-            }
-            if (dealParams.getIexecSecrets() == null) {
-                dealParams.setIexecSecrets(Collections.emptyMap());
-            }
-            return dealParams;
-        } catch (IOException e) {
-            //the requester want to execute one task with the whole string
-            return DealParams.builder()
-                    .iexecArgs(params)
-                    .iexecInputFiles(Collections.emptyList())
-                    .iexecDeveloperLoggerEnabled(false)
-                    .iexecSecrets(Collections.emptyMap())
-                    .build();
-        }
-    }
-
     public static ChainDeal parts2ChainDeal(String chainDealId, Tuple9<String, String, BigInteger, String, String, BigInteger, String, String, BigInteger> dealPt1, Tuple6<BigInteger, byte[], String, String, String, String> dealPt2, Tuple6<BigInteger, BigInteger, BigInteger, BigInteger, BigInteger, BigInteger> config, ChainApp app, ChainCategory category, ChainDataset dataset) {
         if (dealPt1 == null || dealPt2 == null || config == null || app == null || category == null) {
             return new ChainDeal();
@@ -109,7 +85,7 @@ public class ChainDeal {
                 .requester(dealPt2.component3())
                 .beneficiary(dealPt2.component4())
                 .callback(dealPt2.component5())
-                .params(stringToDealParams(dealPt2.component6()))
+                .params(DealParams.createFromString(dealPt2.component6()))
                 .chainCategory(category)
                 .startTime(config.component2())
                 .botFirst(config.component3())
--- a/src/main/java/com/iexec/common/chain/DealParams.java
+++ b/src/main/java/com/iexec/common/chain/DealParams.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.*;
 
 import java.util.List;
+import java.util.Map;
 
 @Data
 @NoArgsConstructor
@@ -50,6 +51,10 @@ public class DealParams {
     @JsonProperty("iexec_result_storage_proxy")
     private String iexecResultStorageProxy;
 
+    // Keys are app secrets indices, values are requester secrets indices
+    @JsonProperty("iexec_secrets")
+    private Map<String, String> iexecSecrets;
+
     //Should be set by SDK
     @JsonProperty("iexec_tee_post_compute_image")
     private String iexecTeePostComputeImage;
@@ -57,6 +62,5 @@ public class DealParams {
     //Should be set by SDK
     @JsonProperty("iexec_tee_post_compute_fingerprint")
     private String iexecTeePostComputeFingerprint;
-
 
 }
--- a/src/main/java/com/iexec/common/chain/DealParams.java
+++ b/src/main/java/com/iexec/common/chain/DealParams.java
@@ -17,11 +17,69 @@
 package com.iexec.common.chain;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.*;
+import lombok.extern.slf4j.Slf4j;
 
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.*;
+
+/**
+ * The DealParams class represents runtime configurations provided for tasks computation.
+ * DealParams allow a requester to provide <b>args</b>, <b>input files</b>, <b>secrets</b>,
+ * <b>result storage configuration</b>... Those configurations are written on the blockchain
+ * when a deal is submitted.
+ * <p>
+ * The data written on the blockchain can be:
+ * <ul>
+ * <li> A valid JSON string representing a map with key -&gt; value pairs.
+ *      The keys are mapped to DealParams fields: iexec_args, iexec_input_files, iexec_secrets, iexec_developer_logger,
+ *      iexec_result_encryption, iexec_result_storage_provider, iexec_result_storage_proxy.
+ * <li> A simple string that will be used for <b>iexec_args</b>
+ * </ul>
+ * <p>
+ * The <b>iexec_secrets</b> key -&gt; value map declares which secrets of the requester
+ * can be accessed by a targeted TEE application.
+ * <ul>
+ * <li>A value should hold the secret name of a secret belonging to the requester.
+ *     This secret must be previously pushed on the SMS.
+ * <li>A key associated to a value indicates how the TEE application can access
+ *     that value.
+ * </ul>
+ * <p>
+ * Example:
+ * <ul>
+ * <li>A requester has a secret value "2000-01-01" pushed on the SMS. The name of
+ *     this secret is "my-birthday".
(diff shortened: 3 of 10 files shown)


src/main/java/com/iexec/common/task/TaskDescription.java
```diff
@@ -142,6 +144,11 @@ public static TaskDescription toTaskDescription(String chainTaskId,
                 chainDeal.getParams().getIexecInputFiles() != null) {
             inputFiles = chainDeal.getParams().getIexecInputFiles();
         }
+        Map<String, String> secrets = Map.of();
+        if (chainDeal.getParams() != null &&
```
mcornaton: Testing whether `chainDeal.getParams()` is null there is pretty strange as we use it a bunch of times without any caution later - [like here](https://github.com/iExecBlockchainComputing/iexec-common/blob/6273701ec7bc749f55d6f7cbe5b3c873f27d2aed/src/main/java/com/iexec/common/task/TaskDescription.java#L159).
So we're sure it'll throw an NPE if `chainDeal.getParams()` is null. What about adding a single test at the beginning of the method so that we're sure this value is never null later? BTW, that's already what's done for `chainDeal`.

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
```diff
@@ -0,0 +1,67 @@
...
+        DealParams params = DealParams.builder()
+                .iexecResultEncryption(false)
+                .iexecDeveloperLoggerEnabled(false)
+                .build();
+        Assertions.assertEquals("{}", RequestOrder.toStringParams(params));
+    }
+
+    @Test
+    void shouldSerializeParametersWithEncryptedResult() {
+        for (String storageProxy : List.of(DealParams.DROPBOX_RESULT_STORAGE_PROVIDER, DealParams.IPFS_RESULT_STORAGE_PROVIDER)) {
```
mcornaton: What about a parametrized test there, instead of a `for` loop?

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
```diff
@@ -0,0 +1,67 @@
...
+                .iexecInputFiles(List.of("file1", "file2"))
+                .build();
+        Assertions.assertEquals(expectedData, RequestOrder.toStringParams(params));
+    }
+
+    @Test
+    void shouldSerializeSecrets() {
+        String expectedData = "{\"iexec_secrets\":{\"0\":\"0\"}}";
+        DealParams params = DealParams.builder()
+                .iexecSecrets(Map.of("0", "0"))
```
mcornaton: I'd prefer to have two distinct values. Otherwise, there could be some switch between key and value and we could not notice that.

src/main/java/com/iexec/common/chain/DealParams.java
```diff
@@ -50,6 +51,10 @@
     @JsonProperty("iexec_result_storage_proxy")
     private String iexecResultStorageProxy;
 
+    // Keys are app secrets indices, values are requester secrets indices
```
james-toussaint, line 112:
```suggestion
    /*
    The "iexec_secrets" key -> value map declares which secrets of the requester 
    can be accessed by a targeted TEE application.
    - A value should hold the secret name of a secret belonging to the requester.
    This secret must be previously pushed on the SMS. 
    - A key a associated to a value indicates how the TEE application can access 
    that value.
    
    Example:
    
    - A requester has a secret value "2000-01-01" pushed on the SMS. The name of 
    this secret is "my-birthday".
    - A requester has a secret value "+331234567890" pushed on the SMS. The name 
    of this secret is "my-phone-number".

    The application developer creates a TEE application which allows requesters 
    to :
    1. Share their birthday by affecting their secret to position "2"
    2. Share their phone number by affecting their secret to position "5"

    In order to share birthday and phone number to the TEE application, the 
    requester should declare:
    "iexec_secrets" : { "2": "my-birthday" , "5": "my-phone-number" }
    
    By doing this, the TEE application code will be able to read following 
    environment variables:
    IEXEC_REQUESTER_SECRET_2=2000-01-01
    IEXEC_REQUESTER_SECRET_5=+331234567890
    */
```

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
james-toussaint: Yes, plus a map having more than 1 mapping

src/main/java/com/iexec/common/chain/DealParams.java
jbern0rd: I agree with your suggestions, plus:
* I will try to add it as a javadoc comment
* Maybe the last part should be documented in `iexec-sms,` `iexec-common` has no knowledge of `IEXEC_REQUESTER_SECRET_` prefix

james-toussaint: Yes, I agree about responsibility of the last part. We can either :
- Add a note stating env var prefix is fake and just for overall understandability purposes
- Relocate that part to `com.iexec.common.utils.IexecEnvUtils`

zguesmi: I think it's a good idea to add `IEXEC_REQUESTER_SECRET_` to `IexecEnvUtils`.

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
```diff
@@ -0,0 +1,67 @@
+package com.iexec.common.sdk.order.payload;
+
+import com.iexec.common.chain.DealParams;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+
+class RequestOrderTest {
```
zguesmi, line 7: Can we add a test that serializes all params at once?

src/main/java/com/iexec/common/task/TaskDescription.java
jbern0rd: It would be possible if `chainDeal.getParams()` was the only call which could throw a NPE.
NPE can be thrown by the following calls: `getBotSize()`, `getChainApp()`, `getChainCategory()`, `getChainDataset()`
It is not possible to do a single test at the beginning with those multiple calls.
**I will remove the unnecessary check and add a TODO.**

Later, I would like to investigate how data read from the chain are populated in Object instances.
It would be great to guarantee that `chainDeal.getParams()` can never be null.

What does everyone think of this suggestion ?

zguesmi: Actually `chainDeal.getParams()` cannot be null. See how it is created in the deal [here](https://github.com/iExecBlockchainComputing/iexec-common/blob/63f66080c6ab269897a7d3bd125479b3b6b0eef1/src/main/java/com/iexec/common/chain/IexecHubAbstractService.java#L403). Cf [here](https://github.com/iExecBlockchainComputing/iexec-common/blob/63f66080c6ab269897a7d3bd125479b3b6b0eef1/src/main/java/com/iexec/common/chain/ChainDeal.java#L71).

zguesmi: Also just for information, the deal params field onchain is not guaranteed to be non-null. So the object `DealParams` will be defined but the content of [iexecArgs](https://github.com/iExecBlockchainComputing/iexec-common/blob/63f66080c6ab269897a7d3bd125479b3b6b0eef1/src/main/java/com/iexec/common/chain/ChainDeal.java#L81) could potentially be null which is not necessarily a problem.

jbern0rd: I will take another look to improve my understanding of the code.
Thanks for the help.

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
jbern0rd: Fixed in https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/7107a8db754fd4807bf1ec149afd78fa41506bd1

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
jbern0rd: Fixed in https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/7107a8db754fd4807bf1ec149afd78fa41506bd1

src/main/java/com/iexec/common/chain/ChainDeal.java
```diff
@@ -72,15 +73,19 @@ public static DealParams stringToDealParams(String params) {
...
+            if (dealParams.getIexecSecrets() == null) {
+                dealParams.setIexecSecrets(Map.of());
             }
             return dealParams;
         } catch (IOException e) {
             //the requester want to execute one task with the whole string
             return DealParams.builder()
                     .iexecArgs(params)
-                    .iexecInputFiles(new ArrayList<>())
+                    .iexecInputFiles(List.of())
```
james-toussaint:
```suggestion
                    .iexecInputFiles(Collections.emptyList())
```
?
https://stackoverflow.com/a/39400310

src/main/java/com/iexec/common/chain/ChainDeal.java
```diff
@@ -72,15 +73,19 @@ public static DealParams stringToDealParams(String params) {
...
                     .iexecDeveloperLoggerEnabled(false)
+                    .iexecSecrets(Map.of())
```
james-toussaint:
```suggestion
                    .iexecSecrets(Collections.emptyMap())
```

build.gradle
```diff
@@ -61,9 +61,7 @@ dependencies {
     implementation 'net.jodah:failsafe:2.4.4'
 
     // test
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
-    testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.6.0'
-    testImplementation 'org.mockito:mockito-core:3.6.0'
+    testImplementation 'org.junit.jupiter:junit-jupiter:5.6.0'
```
james-toussaint, line 75: :+1:

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
jbern0rd: Fixed in https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/7107a8db754fd4807bf1ec149afd78fa41506bd1

src/test/java/com/iexec/common/chain/ChainDealTest.java
```diff
@@ -62,26 +63,57 @@
...
+        Assertions.assertNull(deal.getParams());
+        Assertions.assertNull(deal.getPoolOwner());
+        Assertions.assertNull(deal.getPoolPointer());
+        Assertions.assertNull(deal.getPoolPrice());
+        Assertions.assertNull(deal.getRequester());
+        Assertions.assertNull(deal.getSchedulerRewardRatio());
+        Assertions.assertNull(deal.getStartTime());
+        Assertions.assertNull(deal.getTag());
+        Assertions.assertNull(deal.getTrust());
+        Assertions.assertNull(deal.getWorkerStake());
```
zguesmi: You could use `hasAllNullFieldsOrProperties()` from `org.assertj.core.api.Assertions.assertThat`.

src/test/java/com/iexec/common/chain/ChainDealTest.java
```diff
@@ -62,26 +63,57 @@
...
+    }
+
     @Test
     void shouldReadArgsWithoutJson() {
         DealParams params = stringToDealParams(ARGS);
         assertEquals(ARGS, params.getIexecArgs());
         assertEquals(0, params.getIexecInputFiles().size());
+        assertEquals(0,params.getIexecSecrets().size());
```
zguesmi:
```suggestion
        assertEquals(0, params.getIexecSecrets().size());
```

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
```diff
@@ -48,18 +63,47 @@ void shouldSerializeArgs() {
...
         Assertions.assertEquals(expectedData, RequestOrder.toStringParams(params));
     }
 
     @Test
     void shouldSerializeSecrets() {
-        String expectedData = "{\"iexec_secrets\":{\"0\":\"0\"}}";
+        Map<String, String> secrets = Map.of("0", "secretA", "2", "secretD");
+        String serializedSecrets = secrets.entrySet().stream()
+                .map(entry -> "\"" + entry.getKey() + "\":\"" + entry.getValue() + "\"")
+                .collect(Collectors.joining(","));
```
zguesmi: I find this more readable:
```suggestion
        String serializedSecrets = "{\"0\":\"secretA\", \"2\":\"secretD\"}"
```

jbern0rd: Sadly, I don't have much of a choice here.
The keys are unordered in a Map.
Serialized data can thus be `{"0":"secretA","2":"secretD"}` or `{"2":"secretD","0":"secretA"}`
That's why I decided to perform my own serialization with a stream...

zguesmi: I see. For that reason I generally prefer reading `expectedData` in a `JSONObject` and compare its fields.
No problem, we can keep it that way.

mcornaton: What if you provide a `LinkedHashMap` instead? Its order should be predictable.
See https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html.

src/main/java/com/iexec/common/chain/DealParams.java
jbern0rd: I have a question about this.
Do I just add the String definition to `IexecEnvUtils` ?
`public static final String IEXEC_REQUESTER_SECRET_PREFIX = "IEXEC_REQUESTER_SECRET_";`
Do I add `IEXEC_APP_DEVELOPER_SECRET_PREFIX` as well ?

src/main/java/com/iexec/common/chain/ChainDeal.java
jbern0rd: Fixed in https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/c642ed8631c33d38f36a6e45da21850c8272bf56

src/main/java/com/iexec/common/chain/ChainDeal.java
jbern0rd: Fixed in https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/c642ed8631c33d38f36a6e45da21850c8272bf56

src/test/java/com/iexec/common/chain/ChainDealTest.java
jbern0rd: Fixed in https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/86b25b4e02c469cd822bf7e25a2dfd36e10c0e5b

src/test/java/com/iexec/common/chain/ChainDealTest.java
jbern0rd: Fixed in https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/86b25b4e02c469cd822bf7e25a2dfd36e10c0e5b

src/main/java/com/iexec/common/task/TaskDescription.java
jbern0rd: See https://github.com/iExecBlockchainComputing/iexec-common/pull/298/commits/f3b818be8003c042ebf381f70553e7290a43b315

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
jbern0rd: I think what we really care is proving something like:
`stringToDealParams(toStringParams(dealParams)).equals(dealParams)` but is not the same java Object

`ChainDeal#toStringParams` could be refactored to `DealParams#toString`
`RequestOrder#toStringParams` could be refactored to `DealParams#parseString`

To provide the maximum value, I would prefer to do this kind of refactoring instead of trying to check the correct serialization and deserialization of hard-coded json strings.

src/main/java/com/iexec/common/chain/DealParams.java
zguesmi: We can add both yes. Even though the variables will probably be used only in the SMS, it's a good idea to have them all at the same place.

zguesmi (approved):
Well done!

mcornaton (approved):
LGTM! :)

src/main/java/com/iexec/common/chain/DealParams.java
jbern0rd: See 23794c6  and 7040309

src/test/java/com/iexec/common/sdk/order/payload/RequestOrderTest.java
jbern0rd: See 23794c6

zguesmi (approved):
Clean!

src/main/java/com/iexec/common/chain/DealParams.java
james-toussaint: Thanks!

src/main/java/com/iexec/common/chain/DealParams.java
```diff
@@ -58,5 +121,48 @@
...
 
+    /**
+     * Creates an instance from a JSON string representation.
+     * @param paramString data to create a DealParams instance.
+     *                    The value will be parsed if it is a valid JSON string,
+     *                    it will be used as <b>iexecArgs</b> value otherwise.
+     *                    In the second case, every other parameters will have default values.
+     * @return the created instance
+     */
+    public static DealParams createFromString(String paramString) {
```
james-toussaint, line 132: :+1:

src/test/java/com/iexec/common/chain/DealParamsTest.java
```diff
@@ -0,0 +1,189 @@
...
+                .iexecInputFiles(List.of(FILE2, FILE3))
+                .iexecSecrets(Map.of("2", "secretK", "1", "secretD"))
+                .build();
+        DealParams newParams = DealParams.createFromString(params.toJsonString());
+        assertThat(params).isEqualTo(newParams);
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {DealParams.DROPBOX_RESULT_STORAGE_PROVIDER, DealParams.IPFS_RESULT_STORAGE_PROVIDER})
+    void testSerializationWithAllParams(String storageProvider) {
```
james-toussaint, line 172: :+1:

james-toussaint (approved):
Thanks :)

src/main/java/com/iexec/common/chain/DealParams.java
```diff
@@ -17,11 +17,69 @@
...
 import java.util.Map;
 
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.*;
+
+/**
+ * The DealParams class represents runtime configurations provided for tasks computation.
+ * DealParams allow a requester to provide <b>args</b>, <b>input files</b>, <b>secrets</b>,
+ * <b>result storage configuration</b>... Those configurations are written on the blockchain
+ * when a deal is submitted.
+ * <p>
```
mcornaton, line 38: I usually prefer to use `<br>` tag. Don't know what @jeremyjams and @zguesmi prefer.

mcornaton (approved):
Looks great, thanks!

src/main/java/com/iexec/common/chain/DealParams.java
james-toussaint: b stands for bold :)

jbern0rd: I followed this documentation https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html
I like `<p>` which stands for paragraph but `<br>` is fine as well.

james-toussaint: OK comment was on `<p>` ..!

Default Intellij likes `<p>` too:
image

mcornaton: If both of you prefer `<p>`, let's use that tag so we're consistent :)

PR #300, opened 2022-01-05
Upgrade to docker-java@3.2.12

james-toussaint:


build.gradle
```diff
@@ -54,11 +54,15 @@ dependencies {
...
-    implementation 'com.github.docker-java:docker-java-transport-httpclient5:3.2.5'
+    implementation 'com.github.docker-java:docker-java:3.2.12'
+    // Using zerodep until docker client stops blocking after having called
+    // listContainersCmd() over httpclient5
+    implementation group: 'com.github.docker-java', name: 'docker-java-transport-zerodep', version: '3.2.12'
 
     // failsafe
     implementation 'net.jodah:failsafe:2.4.4'
+    //TODO Migrate from Awaitility to Failsafe everywhere
+    implementation group: 'org.awaitility', name: 'awaitility', version: '4.1.1'
```
jbern0rd: I don't see where this dependency is being used.
Why has it been added here ?

build.gradle
```diff
@@ -54,11 +54,15 @@ dependencies {
     implementation 'io.github.openfeign:feign-slf4j:11.0'
 
     // docker
-    implementation 'com.github.docker-java:docker-java:3.2.5'
-    implementation 'com.github.docker-java:docker-java-transport-httpclient5:3.2.5'
+    implementation 'com.github.docker-java:docker-java:3.2.12'
+    // Using zerodep until docker client stops blocking after having called
+    // listContainersCmd() over httpclient5
+    implementation group: 'com.github.docker-java', name: 'docker-java-transport-zerodep', version: '3.2.12'
```
jbern0rd: Shouldn't we stick to a single dependency notation format ?

zguesmi (approved):
Nice!

build.gradle
james-toussaint: Bad copy paste, updated thanks :)

build.gradle
james-toussaint: https://github.com/iExecBlockchainComputing/iexec-common/blob/a5ca5464ae0c5c1a2786ef5a3bacf0845e5d8a09/src/main/java/com/iexec/common/chain/IexecHubAbstractService.java#L26

mcornaton (approved):

build.gradle
jbern0rd: Thanks, I missed this usage

jbern0rd (approved):
Thanks

PR #302, opened 2022-01-11
Update technical debt

mcornaton:


src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
```diff
@@ -913,8 +888,8 @@ public String getTaskResults(String chainTaskId, Integer chainId) {
                     .viewTaskABILegacy(BytesUtils.stringToBytes(chainTaskId)).send());
             String chainDealId = chainTask.getDealid();
             if (!StringUtils.isEmpty(chainDealId)
-                    && BytesUtils.isBytes32(BytesUtils.stringToBytes(chainDealId))
-                    && !BytesUtils.EMPTY_HEXASTRING_64.equals(chainDealId)){
+                    && BytesUtils.isBytes32(chainDealId)
```
james-toussaint: Thanks for the update which is right, how about `isNonZeroedBytes32` ?

src/main/java/com/iexec/common/chain/eip712/EIP712Utils.java
```diff
@@ -72,7 +72,7 @@ static String encodeBigInteger(BigInteger bigInteger) {
     }
 
     static String encodeByteArray(byte[] byteArray) {
-        if (!BytesUtils.isBytes32(byteArray)) {
+        if (!BytesUtils.isBytes32(BytesUtils.bytesToString(byteArray))) {
```
james-toussaint: `BytesUtils.bytesToString(byteArray)` calls `Numeric.toHexString(bytes)`, so it looks we are calling twice the method until the `return`.

I might say checking `bytes != null && bytes.length == BYTES_32_SIZE` in the `if ` would be more appropriate

src/main/java/com/iexec/common/chain/eip712/EIP712Entity.java
```diff
@@ -43,7 +45,7 @@ protected EIP712Entity(EIP712Domain domain, M message) {
 
     @Override
     public HashMap<String, List<TypeParam>> getTypes() {
-        return types;
+        return new HashMap<>(types);
```
jbern0rd, line 44:
```suggestion
        return types.clone();
```
Would it be more readable this way ???

src/main/java/com/iexec/common/chain/eip712/EIP712Entity.java
```diff
@@ -76,4 +78,13 @@ public String signMessage(ECKeyPair ecKeyPair) {
...
     }
 
+    @JsonIgnore
+    public List<TypeParam> getDomainTypeParams() {
+        return new ArrayList<>(types.get(EIP712Domain.primaryType));
+    }
+
+    @JsonIgnore
+    public List<TypeParam> getMessageTypeParams() {
+        return new ArrayList<>(types.get(getPrimaryType()));
```
jbern0rd, line 84:
```suggestion
        return types.get(getPrimaryType()).clone();
```
??

src/main/java/com/iexec/common/chain/eip712/EIP712Entity.java
mcornaton: I prefer to create a new map by hand instead of using the `clone` method. I find it a bit cleaner.

src/main/java/com/iexec/common/chain/IexecHubAbstractService.java
mcornaton: Sure, much cleaner. Thanks!

src/main/java/com/iexec/common/chain/eip712/EIP712Utils.java
mcornaton: Sure, thanks!

src/main/java/com/iexec/common/chain/eip712/EIP712Entity.java
```diff
@@ -43,7 +45,7 @@ protected EIP712Entity(EIP712Domain domain, M message) {
 
     @Override
     public HashMap<String, List<TypeParam>> getTypes() {
```
james-toussaint: We can use map everywhere
(you can add a todo if you prefer)

src/main/java/com/iexec/common/contract/generated/App.java
```diff
@@ -20,6 +20,8 @@
 import java.util.Arrays;
```
james-toussaint: You should not update what's inside /generated

james-toussaint: If you have time you can setup sonar exclusion for such folder https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/

src/main/java/com/iexec/common/result/eip712/Eip712Challenge.java
```diff
@@ -17,69 +17,49 @@
...
- * Upgrade (and move) this class to get this behaviour (PR TODO)
- * or build your own implementation.
- */
-@Deprecated(forRemoval = true)
-@Data
-@Getter
-@NoArgsConstructor
-public class Eip712Challenge {
-
+public class Eip712Challenge extends EIP712Entity<Message> {
```
james-toussaint, line 26: :rocket:!

src/main/java/com/iexec/common/utils/FileHelper.java
```diff
@@ -420,11 +420,7 @@ private static void printFile(File file, int indent, StringBuilder sb) {
     }
 
     private static String getIndentString(int indent) {
-        StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < indent; i++) {
-            sb.append("│   ");
-        }
-        return sb.toString();
+        return "│   ".repeat(indent);
```
james-toussaint, line 423: Nice

src/test/java/com/iexec/common/sdk/order/payload/DatasetOrderTest.java
```diff
@@ -25,10 +25,10 @@
 
 import java.math.BigInteger;
 
-public class DatasetOrderTest {
+class DatasetOrderTest {
 
     @Test
-    public void getDatasetOrder() throws JsonProcessingException {
+    void getDatasetOrder() throws JsonProcessingException {
```
james-toussaint, line 31: Thanks for all!

src/main/java/com/iexec/common/chain/eip712/EIP712Entity.java
mcornaton: Thanks, I wondered whether I could use a `Map` there. That'll make things cleaner.
See https://github.com/iExecBlockchainComputing/iexec-common/pull/302/commits/490d6ad2e45e0a00aa4323007437ac990677a749.

src/main/java/com/iexec/common/utils/FileHelper.java
mcornaton: Yeah, I didn't know it was a thing in Java 11 but that looks way better than the ugly loop :)

src/main/java/com/iexec/common/contract/generated/App.java
mcornaton: Those are reverted.
That'd be great to set Sonar exclusion but we still can't sync our local Sonar settings with SonarQube so it would still annoy me :laughing:

mcornaton: It should be OK in SonarQube:
image

src/main/java/com/iexec/common/result/eip712/Eip712Challenge.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 IEXEC BLOCKCHAIN TECH
+ * Copyright 2022 IEXEC BLOCKCHAIN TECH
```
zguesmi: Let's keep it consistent with other files.

zguesmi (approved):

jbern0rd (approved):

james-toussaint (approved):

PR #304, opened 2022-01-19
Add no args constructors for Eip712 challenges, they are mandatory for feign clients

jbern0rd:
This fix the following problem
```
org.springframework.http.converter.HttpMessageConversionException: Type definition error: [simple type, class com.iexec.common.result.eip712.Eip712Challenge]; nested exception is com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `com.iexec.common.result.eip712.Eip712Challenge` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```


mcornaton (approved):
Thanks!

james-toussaint (approved):
:+1:

PR #305, opened 2022-01-21
Unwrap AssertionErrors when needed

mcornaton:
`AssertionError`s are built in such a way there's sometimes a cause (~99%), sometimes not (~1%). This may cause some tests to fail.
The idea is to always unwrap these exceptions there, so that we're sure what we should test.

### DIFF
--- a/src/test/java/com/iexec/common/utils/ContextualLockRunnerTests.java
+++ b/src/test/java/com/iexec/common/utils/ContextualLockRunnerTests.java
@@ -3,6 +3,7 @@ package com.iexec.common.utils;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


james-toussaint (approved):

src/test/java/com/iexec/common/utils/ContextualLockRunnerTests.java
```diff
@@ -156,9 +152,20 @@ void getWithLockOnParity() {
...
+            IntStream.range(0, NUMBER_OF_THREADS)
+                    .parallel()
+                    .forEach(threadPosition -> action.accept(threadPosition, remainingCallsPerKey));
+        } catch (AssertionError e) {
+            // Sometimes it has a cause, sometimes not.
+            // When there's a cause,
+            // the "Synchronization failed" message is in this cause.
+            // Otherwise, it's in the exception itself.
+            if (e.getCause() != null) {
+                throw (AssertionError)e.getCause();
```
jbern0rd, line 165: Are we sure the cause can always be cast to an `AssertionError` ?
As `assertThatThrownBy` works with `Throwable`, it may be safer to `throw e.getCause()`
```suggestion
                throw e.getCause();
```

mcornaton: I'm 99% sure it's always an `AssertionError`. If we get anything else, that's another issue and I would prefer the test to fail if there's an unknown issue.

jbern0rd: You have a point

jbern0rd (approved):

PR #307, opened 2022-02-04
Feature/add shared feign builder

jbern0rd:


build.gradle
```diff
@@ -61,8 +61,8 @@ dependencies {
 
     implementation 'ch.qos.logback:logback-classic:1.2.3'
     implementation 'io.github.openfeign:feign-okhttp:11.0'
-    implementation 'io.github.openfeign:feign-jackson:11.0'
-    implementation 'io.github.openfeign:feign-slf4j:11.0'
+    api 'io.github.openfeign:feign-jackson:11.7'
```
james-toussaint:
```suggestion
    api "io.github.openfeign:feign-jackson:${openfeignVersion}"
```

build.gradle
```diff
@@ -82,6 +82,10 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter:5.6.0'
     testImplementation 'org.mockito:mockito-junit-jupiter:3.6.0'
     testImplementation 'org.assertj:assertj-core:3.18.1'
+    testImplementation 'org.testcontainers:testcontainers:1.16.3'
```
james-toussaint: Same for single version

src/main/java/com/iexec/common/utils/FeignBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+ * <li>JacksonDecoder with overridden decode method to deal with strings
+ * <li>SLF4J logger enabled with FULL details. Logs are displayed by configuring the log level on a target interface logger.
+ * <li>application/json Content-Type header
+ * </ul>
+ */
+public class FeignBuilder {
+
+    private FeignBuilder() {};
+
+    public static Feign.Builder createBuilder() {
```
james-toussaint:
```suggestion
    public static Feign.Builder createBuilder(Logger.Level loggerLevel) {
```

jbern0rd: I asked myself to do this like this when writing this class.
It brings several interesting design questions:
* What should be the default `Logger.Level` value when calling `createBuilder` ?
* Do we link this to SLF4J log level ? (For example, if slf4j level is `DEBUG,` feign log level is `FULL` by default, if slf4j log level is `INFO`, feign log level is `BASIC` by default)

What do you think ?

james-toussaint: BASIC by default without SLF4J binding is fine

jbern0rd: done in 27bc189

build.gradle
jbern0rd: done in bec9e47

build.gradle
jbern0rd: done in bec9e47

src/test/java/com/iexec/common/utils/FeignBuilderTest.java
```diff
@@ -0,0 +1,112 @@
...
+import static org.mockserver.model.HttpResponse.response;
+
+/**
+ * The purpose of those tests is to check the {@code JacksonEncoder#encode}
+ * and {@code JacksonDecoder#decode} methods are properly overridden.
+ */
+public class FeignBuilderTest {
+
+    @Container
+    private static final MockServerContainer mockServer = new MockServerContainer(DockerImageName.parse("mockserver/mockserver"))
```
mcornaton: New line after `=` so the line is not too long?

mcornaton (approved):

src/main/java/com/iexec/common/utils/FeignBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+ * <li>application/json Content-Type header
+ * </ul>
+ */
+public class FeignBuilder {
+
+    private FeignBuilder() {};
+
+    public static Feign.Builder createBuilder(Logger.Level logLevel) {
+        return Feign.builder()
+                .encoder(new JacksonEncoder() {
```
zguesmi, line 56: Why do we need these encoder and decoder? Could you add a comment to explain the reason?

src/main/java/com/iexec/common/utils/FeignBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+                        if (type == String.class) {
+                            return new StringDecoder().decode(response, type);
+                        } else {
+                            return super.decode(response, type);
+                        }
+                    }
+                })
+                .logger(new Slf4jLogger())
+                .logLevel(logLevel)
+                .requestInterceptor(template -> template.header("Content-Type", "application/json"));
```
zguesmi, line 78: This implicit behavior could be confusing. Are we sure we want it here?

src/test/java/com/iexec/common/utils/FeignBuilderTest.java
zguesmi: Is it possible to use a Java-native mockServer instance? That would make tests lighter.

src/main/java/com/iexec/common/utils/FeignBuilder.java
jbern0rd: Documentation improved with [09297f0](https://github.com/iExecBlockchainComputing/iexec-common/pull/307/commits/09297f087e853ac857a439c0c932f0cee99a483c)

src/test/java/com/iexec/common/utils/FeignBuilderTest.java
jbern0rd: Done in [0a15e8a](https://github.com/iExecBlockchainComputing/iexec-common/pull/307/commits/0a15e8a407d69ccad1176c7009c9b67fb6f3fdea)

gradle.properties
```diff
@@ -1,3 +1,7 @@
 version=5.9.1
 nexusUser=
 nexusPassword=
+
+dockerJavaVersion=3.2.12
```
james-toussaint: I always wanted to define all external versions here but I feel convention is to store them in build.gradle:
https://github.com/ReactiveX/RxJava/blob/47189ac2ffb71304ca2e44c3d5f51eb9a0ea4e86/build.gradle#L15
https://github.com/web3j/web3j/blob/bfe7d12d7e37adad3955db9a8b28bf199dd0acb2/build.gradle#L12

Do you have many samples where gradle.properties is used instead?

src/main/java/com/iexec/common/utils/FeignBuilder.java
james-toussaint: Thanks!

gradle.properties
jbern0rd: I decided to revert this change.

Dependency version definition is kind of an old-style practice.
It does not correspond to where the gradle project is going at the moment.

The answer to where do we define dependency versions is strongly related to how the source repositories are organised (multiple source repositories, all-in-one repository, ...).
This leads to a much broader discussion which may be a bit early in our situation.
I prefer to postpone it and to take the time to survey what is done in the community.
To complete the links you gave, I found those ones:
* https://github.com/mockito/mockito (dependencies applied from another file)
* https://github.com/spring-projects/spring-framework (dependency management in the root project of a gradle multi-project tree)
* https://github.com/open-telemetry/opentelemetry-java (dependencies defined in a plugin, its source code is located in the `buildSrc` folder)
* https://github.com/opensearch-project/OpenSearch (dependencies defined in a plugin, its source code is located in the `buildSrc` folder)

james-toussaint: Nice, thanks for these references, I have rarely seen this approach : https://github.com/mockito/mockito/blob/main/gradle/dependencies.gradle, always interesting.
I agree to come back to re factorization when internal architecture will change :+1:

src/main/java/com/iexec/common/utils/FeignBuilder.java
james-toussaint: I agree, but we currently don't have many use-cases where this behavior would be different.
Maybe you would like a comment on top?

Note: This builder is a default, it can be override by client

build.gradle
```diff
@@ -82,6 +85,9 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter:5.6.0'
     testImplementation 'org.mockito:mockito-junit-jupiter:3.6.0'
     testImplementation 'org.assertj:assertj-core:3.18.1'
+    // mock-server
+    testImplementation "org.mock-server:mockserver-client-java:5.11.2"
+    testImplementation "org.mock-server:mockserver-junit-jupiter:5.11.2"
```
james-toussaint: :+1:

src/test/java/com/iexec/common/utils/FeignBuilderTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.mockserver.socket.PortFactory;
+
+import static org.mockserver.model.HttpRequest.request;
+import static org.mockserver.model.HttpResponse.response;
+
+/**
+ * The purpose of those tests is to check the {@code JacksonEncoder#encode}
+ * and {@code JacksonDecoder#decode} methods are properly overridden.
+ */
+public class FeignBuilderTest {
```
james-toussaint, line 22: :+1:

james-toussaint (approved):
Very nice, thanks!

src/main/java/com/iexec/common/utils/FeignBuilder.java
jbern0rd: Add documentation in 08c0f55 to explain the default configuration

src/main/java/com/iexec/common/utils/FeignBuilder.java
zguesmi: Thanks!

zguesmi (approved):
Very good :+1:

PR #308, opened 2022-02-08
Dynamic SGX driver mode

mcornaton:


src/main/java/com/iexec/common/docker/DockerRunRequest.java
```diff
@@ -72,30 +73,37 @@ public String getStringArgsCmd() {
         return devices != null ? new ArrayList<>(devices) : List.of();
     }
 
-    // override builder's isSgx() & devices() methods
+    public SgxDriverMode getSgxDriverMode() {
+        return sgxDriverMode != null ? sgxDriverMode : SgxDriverMode.NONE;
```
jbern0rd, line 79: If `sgxDriverMode` is null, should we store `SgxDriverMode.NONE` in `sgxDriverMode` ?

src/main/java/com/iexec/common/docker/DockerRunRequest.java
```diff
@@ -72,30 +73,37 @@ public String getStringArgsCmd() {
...
     public static class DockerRunRequestBuilder { 
-        private boolean isSgx;
+        private SgxDriverMode sgxDriverMode;
         private List<Device> devices;
 
         /**
          * Add an SGX device when isSgx is true.
          * 
-         * @param isSgx
+         * @param sgxDriverMode SGX driver mode
```
jbern0rd, line 96: Could you describe in the javadoc comment what happens depending on the value of `sgxDriverMode` ?

src/test/java/com/iexec/common/docker/DockerRunRequestTests.java
```diff
@@ -75,17 +76,17 @@ void shouldAddDeviceThenAddSgxDevice() {
         DockerRunRequest request = DockerRunRequest.builder()
                 .containerName("containerName")
                 .devices(devices) // changed order here
-                .isSgx(true)
+                .sgxDriverMode(SgxDriverMode.LEGACY)
                 .build();
         System.out.println(request);
```
jbern0rd: Should we remove this or replace it with a SLF4J logger call ?

src/test/java/com/iexec/common/docker/DockerRunRequestTests.java
```diff
@@ -51,16 +52,16 @@ void shouldAddSgxDeviceThenAddDevices() {
         devices.add(device2);
         DockerRunRequest request = DockerRunRequest.builder()
                 .containerName("containerName")
-                .isSgx(true) // notice order here
+                .sgxDriverMode(SgxDriverMode.LEGACY) // notice order here
```
jbern0rd, line 42: What of the other modes ? Should we test them as well ?

src/main/java/com/iexec/common/docker/DockerRunRequest.java
```diff
@@ -72,30 +73,37 @@ public String getStringArgsCmd() {
...
-                    SgxUtils.SGX_DEVICE_PATH,
-                    SgxUtils.SGX_DEVICE_PATH);
-            this.devices.add(sgxDevice);
+
+            for (String deviceName : sgxDriverMode.getDevices()) {
+                Device sgxDevice = new Device(
+                        SgxUtils.SGX_CGROUP_PERMISSIONS,
+                        "/dev/" + deviceName,
+                        "/dev/" + deviceName);
+                this.devices.add(sgxDevice);
```
jbern0rd:
```suggestion
                this.devices.add(Device.parse("/dev/" + deviceName));
```
`Device.parse` has the same result has `new Device("rwm", "device", "device")`

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.iexec.common.sgx;
+
+public enum SgxDriverMode {
+    NONE(new String[]{}),
+    LEGACY(new String[]{"isgx"}),
+    NATIVE(new String[]{"sgx_enclave", "sgx_provision"});
```
jbern0rd:
```suggestion
    LEGACY(new String[]{"/dev/isgx"}),
    NATIVE(new String[]{"/dev/sgx_enclave", "/dev/sgx_provision"});
```
I would prefer to see the full device path stored here

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Returns {@literal false} if given {@link SgxDriverMode} is {@literal null}
+     * or {@link SgxDriverMode#NONE}, {@literal true} otherwise.
+     *
+     * @param driverMode {@link SgxDriverMode} object to check.
+     * @return {@literal false} if given {@link SgxDriverMode} is {@literal null}
+     * or {@link SgxDriverMode#NONE}, {@literal true} otherwise.
+     */
+    public static boolean isDriverModeNotNone(SgxDriverMode driverMode) {
+        return driverMode != null && !driverMode.equals(NONE);
```
jbern0rd:
```suggestion
        return driverMode != null && driverMode != NONE;
```
It seems better to use `!=` operator instead of `equals`
https://www.baeldung.com/a-guide-to-java-enums

src/main/java/com/iexec/common/docker/DockerRunRequest.java
mcornaton: We can do so.

src/main/java/com/iexec/common/docker/DockerRunRequest.java
mcornaton: Sure. JavaDoc is currently wrong as it still refers to `isSgx` which has been removed. Let's update it!

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
mcornaton: I'm not sure it would be better. Devices should always be put in `/dev` folder on Linux. 
BTW, we use the name without its path there:
https://github.com/iExecBlockchainComputing/iexec-worker/blob/23a4b347a64d9a23bb8549653e1890d24a5eba6d/src/main/java/com/iexec/worker/sgx/SgxService.java#L110

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
mcornaton: Done in https://github.com/iExecBlockchainComputing/iexec-common/pull/308/commits/24b0c04479f5ce6a5963a0c09882a81f478e7c41.

src/main/java/com/iexec/common/docker/DockerRunRequest.java
mcornaton: Done in https://github.com/iExecBlockchainComputing/iexec-common/pull/308/commits/24b0c04479f5ce6a5963a0c09882a81f478e7c41.

src/test/java/com/iexec/common/docker/DockerRunRequestTests.java
mcornaton: Replaced with Slf4j in https://github.com/iExecBlockchainComputing/iexec-common/pull/308/commits/24b0c04479f5ce6a5963a0c09882a81f478e7c41.

src/main/java/com/iexec/common/docker/DockerRunRequest.java
mcornaton: Done in https://github.com/iExecBlockchainComputing/iexec-common/pull/308/commits/24b0c04479f5ce6a5963a0c09882a81f478e7c41.

src/main/java/com/iexec/common/docker/DockerRunRequest.java
mcornaton: Done in https://github.com/iExecBlockchainComputing/iexec-common/pull/308/commits/24b0c04479f5ce6a5963a0c09882a81f478e7c41.

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
james-toussaint: I tend to agree with JB, behavior would be more explicit with suggested proposal.

mcornaton: I also agree it would make behavior more explicit but it would make [Worker task](https://github.com/iExecBlockchainComputing/iexec-worker/blob/23a4b347a64d9a23bb8549653e1890d24a5eba6d/src/main/java/com/iexec/worker/sgx/SgxService.java#L110) less intuitive. So I'm really not sure it'd be better in the end.

src/main/java/com/iexec/common/docker/DockerRunRequest.java
```diff
@@ -72,39 +74,48 @@ public String getStringArgsCmd() {
...
                 this.devices = new ArrayList<>();
             }
-            Device sgxDevice = new Device(
-                    SgxUtils.SGX_CGROUP_PERMISSIONS,
-                    SgxUtils.SGX_DEVICE_PATH,
-                    SgxUtils.SGX_DEVICE_PATH);
-            this.devices.add(sgxDevice);
+
+            for (String deviceName : sgxDriverMode.getDevices()) {
+                this.devices.add(Device.parse("/dev/" + deviceName));
```
zguesmi: Do we need to manage permissions here?

mcornaton: I don't think so. Old way was to declare them explicitly, new way is to let `Device.parse` manage them - which currently is the same as before (`rwm`).

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
zguesmi: I also think it's better to put the full device path to make things more explicit.
@mcornaton sorry I didn't understand your last comment.

mcornaton: Workers need to know the name of the devices to check SGX is supported. As we know devices are always inside `/dev`, we can simply do the following:
```bash
ls /dev | grep -w {list of devices names}
```
If we now store the full path to the devices in `SgxDriverMode`, this command line will become strange. Of course we could prepare devices names by splitting paths before running the command, but it would not be clean.

mcornaton: The other solution would be to use `Path` in `SgxDriverMode` such as:
```java
public enum SgxDriverMode {
    NONE(),
    LEGACY(Path.of("/dev", "isgx")),
    NATIVE(Path.of("/dev", "sgx_enclave"), Path.of("/dev", "sgx_provision"))
}
```

But that would lead to the following in `SgxService` of Worker:
```java
        final Path[] devices = sgxDriverMode.getDevices();
        // The following will find SGX devices
        // related to selected SGX driver mode in `/dev`.
        final String cmd = String.format("/bin/sh -c 'echo $(ls /dev | grep -w %s)'",
                Arrays.stream(devices)
                        .map(devicePath -> "-e \"" + devicePath.getFileName() + "\"")
                        .collect(Collectors.joining(" ")));
```
We still explicitly specify the `/dev` folder...

src/test/java/com/iexec/common/docker/DockerRunRequestTests.java
mcornaton: Some new tests there: https://github.com/iExecBlockchainComputing/iexec-common/pull/308/commits/cd3695cb9161c2a152638e238c925d94d7e557d4 :)

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
jbern0rd: With full path, couldn't the test become something like:
`bash -c "test -e <device1> && test -e <device2>"` ?
Return value will be 0 if all devices exist, 1 otherwise.

james-toussaint: Same comment, why can't you test like that : https://github.com/scontain/SH/blob/8c3eecc6eda2c22e682e26c492fc1af7a14dca3a/install_sgx_driver.sh#L932 ?

mcornaton: Both of you are right, thanks! That's updated in my last commits :smile:

jbern0rd (approved):
Thanks for this PR

zguesmi (approved):
Looks good :+1:

src/main/java/com/iexec/common/sgx/SgxDriverMode.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.iexec.common.sgx;
+
+public enum SgxDriverMode {
+    NONE(),
+    LEGACY("/dev/isgx"),
+    NATIVE("/dev/sgx_enclave", "/dev/sgx_provision");
```
james-toussaint, line 22: Thanks for your patience!

james-toussaint (approved):
Nice!

PR #309, opened 2022-02-09
Upgrade Web3J dependency

mcornaton:


jbern0rd:
I would prefer to finalize some of the current PRs before validating this one.
Should we finalize a iexec-common version before merging this ?

mcornaton:
Sure, let's put it aside for a moment - it's not a really important thing, just some debt laying around :)

jbern0rd:
> Sure, let's put it aside for a moment - it's not a really important thing, just some debt laying around :)

I am not sure that this upgrade is not important. Web3j releases between 4.7.0 and 4.9.0 mention EIP712, EIP1559, london fork...
@zguesmi and @jeremyjams should have a better idea of the 'risks' if some exist.
If this upgrade is safe for them, let's do it.
I just try to anticipate the risks and if there could be hidden impacts (on performance for example).

PR #315, opened 2022-03-23
Image pull timeout

mcornaton:


src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -261,17 +262,29 @@ public synchronized boolean removeNetwork(String networkName) {
...
 
     /**
      * Pull docker image and timeout after 1 minute.
      * 
-     * @param imageName
+     * @param imageName Name of the image to pull
      * @return true if image is pulled successfully,
      * false otherwise.
      */
     public boolean pullImage(String imageName) {
```
zguesmi, line 275: If we keep this method we can add the duration of 1 minute in javadoc.

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
```diff
@@ -411,6 +413,14 @@ public void shouldPullImage() {
         dockerClientInstance.removeImage(ALPINE_LATEST);
     }
 
+    @Test
```
zguesmi, line 416: There is no test for successful pulling of an image using the new method.

zguesmi (approved):

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
mcornaton: Where would you like me to add it? It is already in `pullImage` JavaDoc:
```
Pull docker image and timeout after 1 minute.
```

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
mcornaton: True. This is indirectly tested by `shouldPullImage` but let's add another explicit test :)
See https://github.com/iExecBlockchainComputing/iexec-common/pull/315/commits/40dd9019f961e95a853ade189b5de5671e104039.

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
zguesmi: Oh I totally missed, my bad!

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
```diff
@@ -411,6 +413,21 @@ public void shouldPullImage() {
...
+    public void shouldPullImageWithExplicitTimeout() {
+        assertThat(dockerClientInstance.pullImage(ALPINE_LATEST, Duration.of(1, ChronoUnit.MINUTES))).isTrue();
+        assertThat(dockerClientInstance.isImagePresent(ALPINE_LATEST)).isTrue();
+        dockerClientInstance.removeImage(ALPINE_LATEST);
```
jbern0rd, line 420: I think the image should be removed at the beginning of test.
Nothing guarantees it exists or not (could be pulled from integration tests) which could lead to discrepancies in executions.

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
```diff
@@ -411,6 +413,21 @@ public void shouldPullImage() {
         dockerClientInstance.removeImage(ALPINE_LATEST);
     }
 
+    @Test
+    public void shouldPullImageWithExplicitTimeout() {
+        assertThat(dockerClientInstance.pullImage(ALPINE_LATEST, Duration.of(1, ChronoUnit.MINUTES))).isTrue();
```
jbern0rd: We may want to use a longer timeout to avoid unwanted failures due to slow network or loaded host.

jbern0rd (commented):
I have a global comment for hotfixes.
I think we should target minimalistic code change and avoid cosmetic changes specifically for them.
This is to improve our chances to avoid a merge conflict when merging on develop later on.

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
mcornaton: As this is a UT, I don't think we'd like to use a longer timeout. Or we should not always launch this test.

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
mcornaton: I tend to agree with you. However, they are all pulled at the beginning of each test: https://github.com/iExecBlockchainComputing/iexec-common/blob/40dd9019f961e95a853ade189b5de5671e104039/src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java#L99
I'm not really sure why.

jbern0rd: I disagree.
The method is annotated with `@BeforeAll`.
Images are pulled once before the execution of the first test.

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
jbern0rd: Sorry, I will never call a test pulling an image from dockerhub a unit test.
Those tests are integration tests, maybe at the wrong place, but they are integration tests.
If we want unit tests, they need to be entirely rewritten and it is a great undertaking :-)

mcornaton: I totally agree with you. I call them UT because they are started with all other UT's, but they are definitely not.

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
mcornaton: Good catch! Confusion with the other `@BeforeEach` annotated method.

jbern0rd: I don't know how, but I think we need to guarantee 2 things:
* Tests start with a predictable state
* The test implementation (timeout values) for tests are in such a way that we guarantee the tests will succeed (we want to avoid infrastructure side effects because other processes run on test machines)

That's why I suggested to have a big timeout because we don't want to measure the speed of the download, we want to ensure that if the download happens (even if slow), the returned result is `true`.

james-toussaint: https://hub.docker.com/_/alpine?tab=tags
image
I'm fine with 1min, but do you prefer 2min or else?

jbern0rd: I think we need a security factor for the test, just to avoid false positive and side-effects due to the infrastructure being used to do something else at the same time.
I would even put a timeout around 3 minutes, just in case.

mcornaton: Ok, updated to 3 minutes. See https://github.com/iExecBlockchainComputing/iexec-common/pull/315/commits/20d2d531bf6f58fd837eb09c07039654f266b3ba.

jbern0rd: Thank you for the update

jbern0rd (approved):

james-toussaint (approved):

PR #318, opened 2022-04-12
Add `getTaggedWorkerpoolOrders` methods

mcornaton:


src/main/java/com/iexec/common/sdk/marketplace/Marketplace.java
```diff
@@ -31,6 +31,12 @@ public interface Marketplace {
 
     public List<WorkerpoolMarketOrder> getWorkerpoolOrders(String workerpoolAddress);
 
+    List<WorkerpoolMarketOrder> getTaggedWorkerpoolOrders(String tag);
```
james-toussaint:
```suggestion
    List<WorkerpoolMarketOrder> getWorkerpoolOrders(String tag);
```
or 
```suggestion
    List<WorkerpoolMarketOrder> getWorkerpoolOrdersByTag(String tag);
```

zguesmi (approved):

src/main/java/com/iexec/common/sdk/marketplace/MarketplaceClient.java
```diff
@@ -78,6 +78,32 @@ public List<WorkerpoolMarketOrder> getWorkerpoolOrders(String workerpoolAddress)
...
+    public List<WorkerpoolMarketOrder> getTaggedWorkerpoolOrders(String tag) {
+        final MPQueryParams queryParams = MPQueryParams.builder()
+                .minTag(tag)
+                .maxTag(tag)
+                .build();
+        return getOrders(OrderType.WORKERPOOL, queryParams);
+    }
+
+    @Override
+    public Optional<WorkerpoolMarketOrder> getOneTaggedWorkerpoolOrder(String workerpoolAddress, String tag) {
```
jbern0rd: It is really minor, but shouldn't it be better to only query the marketplace for a single order instead of fetching several orders to only keep one ?

src/main/java/com/iexec/common/sdk/marketplace/Marketplace.java
jbern0rd: I prefer the second suggestion.

jbern0rd (commented):
Why do we add 3 methods to only use one in iexec-dataset-api ?

src/main/java/com/iexec/common/sdk/marketplace/MarketplaceClient.java
mcornaton: I don't think the marketplace exposes an endpoint to fetch a single order.

mcornaton:
> Why do we add 3 methods to only use one in iexec-dataset-api ?

As these 3 methods already exist for non-tagged WP orders, I think it's better to add the same methods for tagged WP orders. This could be useful later on.

src/main/java/com/iexec/common/sdk/marketplace/Marketplace.java
mcornaton: Thanks, second suggestion implemented in https://github.com/iExecBlockchainComputing/iexec-common/pull/318/commits/11bc56acf27c534250addf237cbc2ba669c2e989.

src/main/java/com/iexec/common/sdk/marketplace/MarketplaceClient.java
james-toussaint: No, it isn't

src/main/java/com/iexec/common/sdk/marketplace/Marketplace.java
james-toussaint: Suggestion #2 is not consistent with 
`public List<WorkerpoolMarketOrder> getWorkerpoolOrders(String workerpoolAddress);`
We shall upgrade one or the other

src/main/java/com/iexec/common/sdk/marketplace/MarketplaceClient.java
jbern0rd: Could you explain why ?

james-toussaint: https://github.com/iExecBlockchainComputing/iexec-market-api/blob/c2ec2d1ad44b50a487ce77bac0d41afbbc847502/api/src/controllers/router.js#L420

src/main/java/com/iexec/common/sdk/marketplace/Marketplace.java
mcornaton: You're right. Let's add an alias to `getWorkerpoolOrders` and deprecate this current method.

mcornaton: Done in https://github.com/iExecBlockchainComputing/iexec-common/pull/318/commits/eed86f4c2828c72a4e181ba2e57f62b1d19bcc6a.

james-toussaint:
I don't see version bump

mcornaton:
> I don't see version bump

Fixed :)

james-toussaint (approved):

src/main/java/com/iexec/common/sdk/marketplace/MarketplaceClient.java
jbern0rd: I get it, your answers were not about knowing if it's better or not, it is just not possible with the current API which returns a paged result with fixed parameters.
In Spring, there is a really elegant implementation which allows to dynamically size a request with a `PageRequest`. It is possible to configure paging and sorting parameters before a JPA request execution.

jbern0rd: https://www.baeldung.com/spring-data-jpa-pagination-sorting

james-toussaint: Yes, I checked that kind of behavior but page here is just an int right now, not a object :( !

PR #320, opened 2022-04-13
Bugfix/improve serialization deserialization

jbern0rd:


src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
```diff
@@ -59,13 +57,16 @@ public ReplicateStatusUpdate(ReplicateStatus status,
...
         this.details = details;
 
-        this.isSuccess = ReplicateStatus.isSuccess(status);
-
         if (modifier != null && modifier.equals(POOL_MANAGER)) {
            this.date = new Date();
         }
     }
 
+    @JsonIgnore
```
james-toussaint: Nice :)
Why not removing `@JsonIgnore` for Pierre?

jbern0rd: I'm not sure to understand what you propose.
I had to add this annotation because for every method prefixed with `is` or `get`, there seem to be an assumption of an associated field.
If there is no such field, you need to tell it to the mapper through the `JsonIgnore` annotation.

james-toussaint: In order to satisfy Pierre's need, what about moving from `isSuccess()` to `success()`?

jbern0rd: I knew I didn't understand. I will try this and see if it allows to remove the annotation.

jbern0rd: Following this morning discussion and task feedback specifications, see cc3569e

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
```diff
@@ -59,13 +59,21 @@ public ReplicateStatusUpdate(ReplicateStatus status,
...
-
         if (modifier != null && modifier.equals(POOL_MANAGER)) {
            this.date = new Date();
         }
     }
 
+    /**
+     * Returns {@link ReplicateStatus#isSuccess(ReplicateStatus)} result.
+     * The result will be serialized in json and appear as {@code success} field.
+     * @return false if {@code status} is a failure status, true otherwise.
```
james-toussaint:
```suggestion
     * @return true if {@code status} is a successful status, false otherwise.
```

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
```diff
@@ -16,29 +16,29 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-import static com.iexec.common.replicate.ReplicateStatusModifier.*;
+import java.util.Date;
+
+import static com.iexec.common.replicate.ReplicateStatusModifier.POOL_MANAGER;
```
james-toussaint, line 28:
```suggestion
import static com.iexec.common.replicate.ReplicateStatusModifier.*;
```
?

src/test/java/com/iexec/common/replicate/ReplicateStatusUpdateTests.java
```diff
@@ -0,0 +1,45 @@
...
+public class ReplicateStatusUpdateTests {
+
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    @Test
+    void shouldSerializeAndDeserializeReplicateStatusUpdate() throws JsonProcessingException {
+        ReplicateStatusUpdate statusUpdate = ReplicateStatusUpdate.poolManagerRequest(CREATED);
+        Date date = statusUpdate.getDate();
+        String jsonString = mapper.writeValueAsString(statusUpdate);
+        String expectedString = "{\"status\":\"CREATED\",\"modifier\":\"POOL_MANAGER\""
```
james-toussaint: Minor, something like
```suggestion
        String expectedString = "{"
        + "\"status\":\"CREATED\","
        + "\"modifier\":\"POOL_MANAGER\,""
```
?

james-toussaint (approved):
Thanks :)

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
jbern0rd: For those imports, I tend to let IDE decide.
An implicit rule seems to be something like:
if there are less than 5 imports, import should be explicitly defined and wildcards should be avoided.

src/test/java/com/iexec/common/replicate/ReplicateStatusUpdateTests.java
jbern0rd: See a364ff8

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
jbern0rd: See a364ff8

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
james-toussaint: I agree, keep the default, we'll re-talk about that if I found we don't have same defaults

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
james-toussaint: Thanks

PR #324, opened 2022-05-03
Enable default feign Retryer in FeignBuilder

jbern0rd:


src/main/java/com/iexec/common/utils/FeignBuilder.java
```diff
@@ -82,7 +84,8 @@ public Object decode(Response response, Type type) throws IOException {
                 })
                 .logger(new Slf4jLogger())
                 .logLevel(logLevel)
-                .requestInterceptor(template -> template.header("Content-Type", "application/json"));
+                .requestInterceptor(template -> template.header("Content-Type", "application/json"))
```
james-toussaint, line 87: What do you think about starting with:
`public static Feign.Builder createBuilder(Logger.Level logLevel, boolean retrier)`

jeremyjams, line 87: What do you think about starting with:
`public static Feign.Builder createBuilder(Logger.Level logLevel, boolean retrier)`

mcornaton (approved):
LGTM, thanks :)

src/main/java/com/iexec/common/utils/FeignBuilder.java
jbern0rd: I don't think this is a good idea.
If the retryer needs to be optional, I prefer to drop this merge request.
As we return a Feign.Builder, the user will be able to choose to enable a retryer directly when building a feign client instance.

On a side note, I almost add to add the retryer for all the libraries.
That's why I decided to add it in the root builder.

To help on a decision, we need to answer the following questions :
* When do we need the retryer ?
* When must a retryer not be defined ?
* Is there a problem to always define a default retryer ?

james-toussaint: I get your points, my main concern was introducing retriers on locations we don't need, then in the future, think that these (now used) retriers are required, and be "scared" to remove them

jbern0rd: No problem, this feature will thus be closed and won't be merged.
It will be the responsibility of a client builder to know if a retryer is needed or not.
Thanks for your reviews.

PR #326, opened 2022-05-04
Feature/task feedback

mcornaton:
- Upgrade `ApiResponseBody`
- Add `ApiResponseBodyDecoder` to decode body of response as String
- Add `teeSessionGenerationError` to `ReplicateStatusDetails`


src/main/java/com/iexec/common/tee/TeeSessionGenerationError.java
```diff
@@ -0,0 +1,56 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.iexec.common.tee;
+
+public enum TeeSessionGenerationError {
```
james-toussaint: Can you remind me/explain why `TeeSessionGenerationError` isn't exposed by the sms-library itself?

src/main/java/com/iexec/common/web/ApiResponseBody.java
```diff
@@ -16,20 +16,40 @@
...
 @NoArgsConstructor
 @AllArgsConstructor
 @JsonInclude(JsonInclude.Include.NON_NULL)
-public class ApiResponseBody<T> {
-    private List<String> errors;
-    private T data;
-}
+public class ApiResponseBody<D, E> {
+    private D data;
+    private E errors;
```
james-toussaint:
```suggestion
    private E error;
```

src/main/java/com/iexec/common/tee/TeeSessionGenerationError.java
mcornaton: It is needed in `ReplicateStatusDetails` which is part of Common.

james-toussaint: Thanks :)
Do you think 
```
ReplicateStatusDetails{
  private String teeSessionGenerationError;
}
```
might be acceptable since for current DevX purposes?

src/main/java/com/iexec/common/tee/TeeSessionGenerationError.java
```diff
@@ -0,0 +1,56 @@
...
+    EXECUTION_NOT_AUTHORIZED_INVALID_SIGNATURE,
+
+    GET_TASK_DESCRIPTION_FAILED,
+
+    NO_SESSION_REQUEST,
+    NO_TASK_DESCRIPTION,
+
+    PRE_COMPUTE_GET_DATASET_SECRET_FAILED,
+
+    COMPUTE_NO_ENCLAVE_CONFIG,
```
james-toussaint:
```suggestion
    APP_COMPUTE_NO_ENCLAVE_CONFIG,
```

src/main/java/com/iexec/common/tee/TeeSessionGenerationError.java
mcornaton: Back to our usual debate - enum or String :)
But I think this would be acceptable there. Let's rework this!

jbern0rd: If the Enum works, I prefer we use it.
It adds type safety and more robustness in the system with a field taking value in a finite list of values.
Or did I misunderstand something ?

mcornaton: I partially agree with you @jbern0rd. Having type safety is a great feature but having `TeeSessionGenerationError` in Common looks strange to me. I prefer it to be in the iexec-sms-library project.

src/main/java/com/iexec/common/web/ApiResponseBody.java
mcornaton: See https://github.com/iExecBlockchainComputing/iexec-common/pull/326/commits/067b770f2e74d5c1677dddcff47ed977b9f0ff57.

src/main/java/com/iexec/common/tee/TeeSessionGenerationError.java
mcornaton: This class has been moved to iexec-sms-library, so please see [this commit](https://github.com/iExecBlockchainComputing/iexec-sms/pull/93/commits/be90c2d1d94d4deba6c1a9b167d24a6a249ed5c6).

jbern0rd (approved):

src/main/java/com/iexec/common/web/ApiResponseBody.java
james-toussaint: Thanks!

src/main/java/com/iexec/common/tee/TeeSessionGenerationError.java
james-toussaint: :pray:

src/main/java/com/iexec/common/web/ApiResponseBodyDecoder.java
```diff
@@ -0,0 +1,53 @@
...
+        JavaType javaType = mapper.getTypeFactory().constructParametricType(ApiResponseBody.class, dataClass, errorClass);
+        try {
+            return Optional.ofNullable(mapper.readValue(response, javaType));
+        } catch (JsonProcessingException e) {
+            log.warn("Failed to decode response", e);
+            return Optional.empty();
+        }
+    }
+
+    public static <D,E> Optional<D> getDataFromResponse(String response, Class<D> dataClass, Class<E> errorClass) {
```
james-toussaint: Would you be able to achieve something like that:?
```suggestion
    public static <D> Optional<D> getDataFromResponse(String response, Class<D> dataClass) {
            return decodeResponse(response, dataClass, Object.class)
                .map(ApiResponseBody::getData);
```

mcornaton: Good idea, see https://github.com/iExecBlockchainComputing/iexec-common/pull/326/commits/6bcba1326309ebaed7a2ff6c489c7724bfd29d6a.

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
```diff
@@ -76,11 +76,13 @@ public boolean isSuccess() {
 
     @JsonIgnore
     public ReplicateStatusDetails getDetailsWithoutStdout() {
```
james-toussaint: `public ReplicateStatusDetails getDetailsWithoutLogs()`
?
The idea is to avoid printing `details.toString()` in a small `[details:{}]`

src/main/java/com/iexec/common/replicate/ReplicateStatusDetails.java
```diff
@@ -36,8 +36,7 @@ public class ReplicateStatusDetails {
     private String chainCallbackData;
     private String errorMessage;
     private ReplicateStatusCause cause;
-    private String stdout;
-    private String stderr;
+    private ReplicateLogs replicateLogs;
```
james-toussaint: Very good!
I would renaming field (`replicate[...]` might be useless since we are in a `ReplicateDetails`) and eventually would change type too.
`private Logs appLogs;` or `private ComputeLogs appLogs;`
?

src/main/java/com/iexec/common/replicate/ReplicateStatusUpdate.java
mcornaton: Sure, I was working on it. See https://github.com/iExecBlockchainComputing/iexec-common/pull/326/commits/d4e127fdd7a39e0103bbb6568a804fd6e6c2f3bb :)

src/main/java/com/iexec/common/replicate/ReplicateStatusDetails.java
mcornaton: What about `AppComputeLogs`? This would be similar to what we already have in the SMS.

james-toussaint: Yes, that would be fine (better than replicate logs I think), but I'm scared someone asks for preComputeLogs and postComputeLogs. If so, we would need to re-update class name to avoid duplicates. Knowing this, what do you prefer?

mcornaton: I see your point. Let's use `ComputeLogs`.

james-toussaint: We already heave `DockerLogs`, but it seems better to have a distinct `ComputeLogs` for independant presentation layer.

mcornaton: Please see https://github.com/iExecBlockchainComputing/iexec-common/pull/326/commits/2189109bf8fc50036f9605d2d8c8d22f6df18bda :)

src/main/java/com/iexec/common/replicate/ReplicateStatusDetails.java
```diff
@@ -29,23 +29,32 @@
 public class ReplicateStatusDetails {
 
     public static final int MAX_STDOUT_LENGTH = 100000;
+    public static final int MAX_STDERR_LENGTH = 100000;
```
james-toussaint: Using a single `MAX_STD_LENGTH`, should be fine, what do you think?
If you agree, you can use a single method for substring stds

src/main/java/com/iexec/common/replicate/ComputeLogs.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
```
james-toussaint, line 29: `NON_EMPTY` or `NON_NULL`?

jbern0rd: `NON_NULL` seems to be included in `NON_EMPTY`.
The later seems to cover as well empty Optional which would return `null`.
From https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JsonInclude.Include.html
In my understanding, it should be something like
NON_EMPTY = NON_NULL && NON_ABSENT

src/main/java/com/iexec/common/replicate/ReplicateStatusDetails.java
mcornaton: Sure, we can do so. Please see https://github.com/iExecBlockchainComputing/iexec-common/pull/326/commits/56f5f3b8b7df6dcd848759bba01d0fcd0325d93e.

src/main/java/com/iexec/common/replicate/ReplicateActionResponse.java
```diff
@@ -72,15 +72,15 @@ public static ReplicateActionResponse failure(ReplicateStatusCause cause) {
 
     public static ReplicateActionResponse failureWithStdout(String stdout) {
```
james-toussaint, line 73: `@Deprecated`
and/or
`failureWithLogs`
?

mcornaton: We still use it for Post-Compute run. Should we also include stderr of Post-Compute?

src/main/java/com/iexec/common/replicate/ReplicateStatusDetails.java
```diff
@@ -27,25 +27,30 @@
...
+    private String teeSessionGenerationError; // null means unset
 
     public ReplicateStatusDetails(ReplicateStatusDetails details) {
         chainReceipt = details.getChainReceipt();
         resultLink = details.getResultLink();
         chainCallbackData = details.getChainCallbackData();
         errorMessage = details.getErrorMessage();
         cause = details.getCause();
-        stdout = details.getStdout();
+        computeLogs = details.getComputeLogs() == null
```
james-toussaint, line 51: Can you explain why not
`computeLogs = details.getComputeLogs();`
directly?

mcornaton: Because if we'd call `tailLogs` on either the `ReplicateStatusDetails` base or the copy, it would also tails on the other one. So duplicating there seems mandatory.

src/main/java/com/iexec/common/replicate/ReplicateActionResponse.java
james-toussaint: Ok, let's keep it that way for now

james-toussaint (approved):

PR #328, opened 2022-05-17
Feature/improve eip712 challenge implementation

jbern0rd:


src/main/java/com/iexec/common/result/eip712/Eip712ChallengeUtils.java
```diff
@@ -25,6 +25,10 @@
 
 import java.math.BigInteger;
 
+/**
+ * @deprecated To be removed
+ */
+@Deprecated
```
zguesmi:
```suggestion
@Deprecated(forRemoval = true)
```

src/main/java/com/iexec/common/chain/eip712/TypeParam.java
```diff
@@ -28,4 +34,12 @@ public class TypeParam {
     private String name;
     private String type;
 
+    /**
+     * Gets the description of a single parameter as a String.
+     * @return The {@code "type name"} string
```
zguesmi: How about
```suggestion
     * @return The {@code "<type> <name>"} string
```
to highlight that both parts are dynamic.

src/main/java/com/iexec/common/chain/eip712/EIP712Challenge.java
```diff
@@ -0,0 +1,86 @@
...
+        );
+    }
+
+    /**
+     * Represents the instance as a JSON string that will be exchanged between services.
+     * @return the JSON string representing the instance
+     */
+    public String toJsonString() {
+        ObjectMapper mapper = new ObjectMapper();
+        Arrays.asList(NON_NULL, NON_EMPTY, NON_DEFAULT).forEach(mapper::setSerializationInclusion);
```
mcornaton: I think `NON_NULL` is included in `NON_EMPTY`.

src/main/java/com/iexec/common/chain/eip712/EIP712Challenge.java
```diff
@@ -0,0 +1,86 @@
...
+        try {
+            return mapper.writeValueAsString(this);
+        } catch (JsonProcessingException e) {
+            log.error("EIP-712 challenge serialization failed.", e);
+            return "";
```
mcornaton: Shouldn't this re-throw the exception instead of returning an empty String?

src/test/java/com/iexec/common/chain/eip712/EIP712ChallengeTests.java
```diff
@@ -0,0 +1,111 @@
...
+                        "{\"name\":\"name\",\"type\":\"string\"}," +
+                        "{\"name\":\"version\",\"type\":\"string\"}," +
+                        "{\"name\":\"chainId\",\"type\":\"uint256\"}" +
+                        "]," +
+                        "\"Challenge\":[{\"name\":\"challenge\",\"type\":\"string\"}]}," +
+                        "\"domain\":{\"name\":\"COMMON\",\"version\":\"1\",\"chainId\":15}," +
+                        "\"message\":{\"challenge\":\"challenge\"}," +
+                        "\"primaryType\":\"Challenge\"}");
+        String domainType = "EIP712Domain(string name,string version,uint256 chainId)";
+        String domainSeparator = HashUtils.concatenateAndHash(
```
mcornaton: Not sure whether it should be dynamically computed or static. Usually, when writing unit tests, I think we should not rely on production code as it may broke at some point.

mcornaton (commented):
This looks better now, thanks!

src/main/java/com/iexec/common/result/eip712/Eip712ChallengeUtils.java
jbern0rd: done in 40711e6

src/test/java/com/iexec/common/chain/eip712/EIP712ChallengeTests.java
jbern0rd: I am not sure to see your point.
Here, `domainSeparator` is computed and its value is compared against `eip712Challenge.getDomain().getDomainSeparator()` as well as a static value with 2 calls to `isEqualTo`.
It is a kind of 3-way comparison to validate that our class implement the correct algorithm and produce the correct value.
What would you do differently ?

src/main/java/com/iexec/common/chain/eip712/EIP712Challenge.java
jbern0rd: sure, just fixed in 88a5ce3

src/main/java/com/iexec/common/chain/eip712/TypeParam.java
jbern0rd: done in 843c797

src/test/java/com/iexec/common/chain/eip712/EIP712ChallengeTests.java
mcornaton: OK, I see what you mean. I had not read the last line of the test, where you check those values against a static value. So it's alright, sorry :)

src/main/java/com/iexec/common/chain/eip712/EIP712Challenge.java
jbern0rd: better, the method could simply be removed.
I anticipated I would need it, but that wasn't the case.
Removed in 4a6eff7

src/test/java/com/iexec/common/chain/eip712/EIP712ChallengeTests.java
jbern0rd: No problem, resolving this thread

mcornaton (approved):
:+1:

src/main/java/com/iexec/common/chain/eip712/EIP712Challenge.java
mcornaton: Thanks!

zguesmi (approved):

PR #330, opened 2022-05-20
Bugfix/eip712 serialization and deserialization

jbern0rd:


src/main/java/com/iexec/common/chain/eip712/EIP712Domain.java
```diff
@@ -64,21 +62,22 @@ public EIP712Domain(String name, String version, long chainId, String verifyingC
...
         this.chainId = chainId;
         this.verifyingContract = verifyingContract;
-
-        this.initTypes();
     }
 
-    private void initTypes() {
-        this.types.addAll(Arrays.asList(
+    @JsonIgnore
+    public List<TypeParam> getTypes() {
```
mcornaton, line 68: Not sure whether `types` are used, but this change would lead to recomputing `types` value each time it is needed. Isn't it an issue? It is used a single time?

mcornaton (approved):

src/main/java/com/iexec/common/chain/eip712/EIP712Domain.java
```diff
@@ -64,21 +62,22 @@ public EIP712Domain(String name, String version, long chainId, String verifyingC
         this.version = version;
         this.chainId = chainId;
         this.verifyingContract = verifyingContract;
-
-        this.initTypes();
```
zguesmi, line 68: Why not keeping both methods, `initTypes` and `getTypes\ ?

zguesmi (approved):

src/main/java/com/iexec/common/chain/eip712/EIP712Domain.java
jbern0rd: It is the crux of the problem.
The `initTypes` method was called from the constructors, including the no args one.
Thus, when deserializing, we always defined the types as a list without `verifyingContract` because deserialization calls the empty constructor before writing the fields values.
At the same time, the types list is only used when creating an EIP712Entity instance from a correctly built EIP712Domain instance.
The easier solution was to remove the `types` field and the `initTypes` method and to implement a `getTypes` method to return the correct list when called from EIP712Entity.

PR #336, opened 2022-06-14
Add timeout information on pre/post compute

mcornaton:


src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
```diff
@@ -150,6 +156,7 @@ void shouldCreateUnauthenticatedClientWithCustomRegistry() {
         assertThat(instance.getClient().authConfig().getPassword()).isNull();
     }
 
+    @Disabled
```
jbern0rd: Involuntary commit ?

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
```diff
@@ -474,6 +481,7 @@ void shouldNotPullImageSinceDockerCmdException() {
      * Try to pull a private image from iexechub, require valid login and permissions.
      * The test will fail if Docker Hub credentials are missing or invalid.
      */
+    @Disabled
```
jbern0rd: Involuntary commit ?

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
mcornaton: Good catch, thanks! See https://github.com/iExecBlockchainComputing/iexec-common/pull/336/commits/a6fbbd4cde027d4296ac1823c92df1961defb76c :)

src/test/java/com/iexec/common/docker/client/DockerClientInstanceTests.java
mcornaton: Good catch, thanks! See https://github.com/iExecBlockchainComputing/iexec-common/pull/336/commits/a6fbbd4cde027d4296ac1823c92df1961defb76c :)

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -411,27 +413,41 @@ public DockerRunResponse run(DockerRunRequest dockerRunRequest) {
...
+            log.info("Finished running docker container [name:{}, isSuccessful:{}]",
+                    containerName, isSuccessful);
+            dockerRunResponse.setFinalStatus(
+                    isSuccessful
+                            ? DockerRunFinalStatus.SUCCESS
+                            : DockerRunFinalStatus.FAILED);
+        } catch (TimeoutException e) {
+            log.error(e.getMessage());
+            dockerRunResponse.setFinalStatus(DockerRunFinalStatus.TIMEOUT);
+            if (!stopContainer(containerName)) {
```
jbern0rd, line 436: Something doesn't seem alright to me in this method.
I understand we retrieve logs then exit if `stopContainer` does not behave properly.
But with the current implementation, we return before calling the `removeContainer` method which happens during a normal execution.
Shouldn't we log the error and try to kill the container, then try to follow through the end of the code ?

jbern0rd: Don't we need a `stopContainer` method where we use this ?
Not-null timeout to have a grace period before a forceful kill
https://github.com/docker-java/docker-java/blob/master/docker-java-api/src/main/java/com/github/dockerjava/api/command/StopContainerCmd.java#L15

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -411,27 +413,41 @@ public DockerRunResponse run(DockerRunRequest dockerRunRequest) {
...
+                return dockerRunResponse;
+            }
+        } catch (IllegalArgumentException e) {
+            log.error(e.getMessage());
         }
+
         getContainerLogs(containerName).ifPresent(dockerRunResponse::setDockerLogs);
         if (!removeContainer(containerName)) {
             log.error("Failed to remove container after run [name:{}]", containerName);
+            dockerRunResponse.setFinalStatus(DockerRunFinalStatus.FAILED);
```
james-toussaint: Not sure to get that one :thinking: 
In the case of a timeout, even if the the container wasn't removed, shouldn't the final status be `TIMEOUT`?

jbern0rd: A the code is written, we will never try to remove the container if there is a timeout.
Your comment is related to mine above and points another trap to avoid.

mcornaton: TBH, not sure about that one, too. Current implementation always returns a failure (`response.isSuccessful() == false`) when removal fails, even though the Docker container successfully exited. If we extend this behavior, even though the Docker container timed out, we should return a failure.
But we can fix this behavior so that it returns a timeout status instead. Let me know what you prefer!

james-toussaint: Generally speaking, to me, stop or remove container should be done in best-effort mode (with just a warn if necessary), but returned response should stay the same, don't you think?

mcornaton: I mostly agree with you. I was simply preserving the current implementation. I'd be happy to fix it!

jbern0rd: By parsing the library, I found that the docker-java StopContainerCmd supports a timeout and can do a forceful kill if the timout is reached.
Shouldn't we leverage this to be sure a container is stopped and avoid polluting the user system ?

james-toussaint: I see your point, let's update that part if you can and then we can attack next point, thank you :)

jbern0rd: https://github.com/docker-java/docker-java/blob/master/docker-java-api/src/main/java/com/github/dockerjava/api/command/StopContainerCmd.java#L14

mcornaton: Thanks @jeremyjams, see https://github.com/iExecBlockchainComputing/iexec-common/pull/336/commits/5116472ce025e83279c2ce2cc5cd91368c1508a6.

mcornaton: @jbern0rd > Do you know how this exactly works? If `timeout` is set to 0, wouldn't that always force kill the container?

jbern0rd: docker-java implementation is a thin wrapper around docker API
https://docs.docker.com/engine/api/v1.41/#operation/ContainerStop
The question is to understand how the docker API works.

jbern0rd (approved):
Thanks for this update

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
jbern0rd: I missed part of the algorithm yesterday.
If the container is not stopped, there is no point in trying to remove it.
If the container is stopped, we continue executing code outside of the try block.

The open point is can the `stopContainer` command fail or will it always succeed ?

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -411,27 +413,39 @@ public DockerRunResponse run(DockerRunRequest dockerRunRequest) {
...
-        int exitCode = waitContainerUntilExitOrTimeout(containerName, timeoutDate);
-        dockerRunResponse.setContainerExitCode(exitCode);
-        boolean isTimeout = exitCode == -1;
-        boolean isSuccessful = !isTimeout && exitCode == 0L;
-        if (isTimeout && !stopContainer(containerName)) {
-            log.error("Failed to force-stop container after timeout [name:{}]", containerName);
-            return dockerRunResponse;
+        boolean isSuccessful;
+        try {
+            int exitCode = waitContainerUntilExitOrTimeout(containerName, timeoutDate);
```
james-toussaint, line 423: Cleaner thanks!

james-toussaint (approved):

PR #343, opened 2022-07-15
Post Gramine session to SPS

mcornaton:
- Add a `TeeEnclaveProvider` field to `TaskDescription`
- Add `TeeUtils#getTeeEnclaveProvider` method
- Update `ReplicateStatusCause` members to match updated `TeeSessionGenerationError`


src/main/java/com/iexec/common/sgx/SgxDriverMode.java
```diff
@@ -19,7 +19,7 @@
 public enum SgxDriverMode {
     NONE(),
     LEGACY("/dev/isgx"),
-    NATIVE("/dev/sgx_enclave", "/dev/sgx_provision");
+    NATIVE("/dev/sgx/enclave", "/dev/sgx/provision");
```
jbern0rd: I don't really like this
I would prefer to keep for as long as possible a system which can run on scone and on gramine.
On my azure vm, freshly installed, it is back to `/dev/sgx_enclave` and `/dev/sgx_provision`.
We need to understand why and when it changes and create a new entry with a new mode imho.

jbern0rd: I found this, but still lack a good bit of context to know where lies the truth
https://github.com/intel/linux-sgx/issues/772

jbern0rd: The machine where developments have been done is an ubuntu 18.04.
Its kernel version is before official sgx driver integration in linux kernel.
New machines are in Ubuntu 20.04 and have the correct driver with the correct devices.
If you agree with the proposal and the understanding, revert to `/dev/sgx_enclave` and `/dev/sgx_provision`

james-toussaint: Yes, totally, we didn't have access to the correct machines/drivers at the time hence the custom setting, this should be removed now :+1:

PR #351, opened 2022-08-12
Run Gramine task on worker

mcornaton:


src/main/java/com/iexec/common/chain/DealParams.java
```diff
@@ -121,6 +121,9 @@ public class DealParams {
     @JsonProperty("iexec_tee_post_compute_fingerprint")
     private String iexecTeePostComputeFingerprint;
 
+    @JsonProperty("iexec_sms_url")
```
zguesmi, line 124: Should we use a "default" sms if the url is not specified in the deal?

zguesmi: It would also reduce gas cost because I think most of the time we would like to use the default SMS.

mcornaton: I don't think we would like to use a default SMS as of now. It may be provided by our SMS catalog in the future, depending on the TEE tag. We'll have to discuss this with @PierreJeanjacquot though.
I propose we stay with no default SMS for now and we'll change it depending on our discussion. Is this OK for you?

zguesmi: No problem yes.

james-toussaint (approved):

jbern0rd (approved):

jbern0rd (commented):
Why do you use symbolic links instead of official SGX device ?
see `ls -Alh /dev/sgx` links pointing to `/dev/sgx_*` devices

src/main/java/com/iexec/common/replicate/ReplicateStatusCause.java
```diff
@@ -79,6 +79,7 @@ public enum ReplicateStatusCause {
     TEE_SESSION_GENERATION_NO_SESSION_REQUEST,
     TEE_SESSION_GENERATION_NO_TASK_DESCRIPTION,
     TEE_SESSION_GENERATION_GET_SESSION_FAILED,
+    TEE_SESSION_GENERATION_NO_SMS_FOR_TASK,
```
jbern0rd: Is it used somewhere ? It does not seem to be the case.

mcornaton: You're right, it is not used anymore. Please see https://github.com/iExecBlockchainComputing/iexec-common/pull/351/commits/2e72b3880e7983fa7fcf28b9d8623299410baba6.

PR #356, opened 2022-09-21
Feature/platform registry

james-toussaint:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=6.0.1
+version=6.0.1-platform
```
james-toussaint:
```suggestion
version=6.0.1
```

src/main/java/com/iexec/common/notification/TaskNotificationExtra.java
```diff
@@ -31,6 +31,7 @@
 public class TaskNotificationExtra {
 
     private WorkerpoolAuthorization workerpoolAuthorization;
+    private String smsUrl;
```
jbern0rd, line 34: There are strong similarities between `TaskNotificationExtra` and `ReplicateTaskSummary`.
There should be a possible improvement of the model here later.

jbern0rd (approved):

mcornaton (approved):

src/main/java/com/iexec/common/notification/TaskNotificationExtra.java
james-toussaint: I agree!

PR #357, opened 2022-10-04
Feature/rename tee enclave provider

jbern0rd:


src/main/java/com/iexec/common/tee/TeeUtils.java
```diff
@@ -99,6 +104,23 @@ public static TeeEnclaveProvider getTeeEnclaveProvider(String hexTag) {
         return null;
     }
 
+    /**
+     * Returns enclave provider matching given {@code hexTag}.
```
mcornaton:
```suggestion
     * Returns framework matching given {@code hexTag}.
```

mcornaton:
Could you please add `@deprecated` JavaDoc tags to point to the new `TeeFramework` enum?

mcornaton (approved):
Thanks! :)

src/main/java/com/iexec/common/tee/TeeEnclaveConfiguration.java
```diff
@@ -30,7 +31,11 @@
 @Builder
 public class TeeEnclaveConfiguration {
 
+    /** @deprecated */
+    @Deprecated(forRemoval = false)
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     private TeeEnclaveProvider provider;
+    private TeeFramework framework;
```
james-toussaint, line 38:
```suggestion
    @JsonInclude(JsonInclude.Include.NON_NULL)
    private TeeFramework framework;
```
?

jbern0rd: I am not sure about this. `TeeFramework` should never be null here.
Unless I'm missing something, a `TeeEnclaveConfiguration` not backed on a TEE framework has no meaning.
I added the annotation on `provider` to cleanup the JSON when core and worker will shift from `provider` to `framework` field.

james-toussaint: Your are right, I was think about `@NonNull` !

james-toussaint (approved):
Thanks :)

PR #361, opened 2022-11-14
Feature/docker rework step 1

jbern0rd:


build.gradle
```diff
@@ -67,7 +68,7 @@ dependencies {
     // json
     implementation 'org.json:json:20201115'
 
-    implementation 'ch.qos.logback:logback-classic:1.2.3'
+    implementation 'ch.qos.logback:logback-classic'
```
james-toussaint, line 71: Can you explain why removing the version?

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
```diff
@@ -164,24 +162,20 @@ public synchronized boolean removeVolume(String volumeName) {
...
-        if (!isVolumePresent(volumeName)) {
-            log.info("No docker volume to remove [name:{}]", volumeName);
-            return false;
-        }
         try (RemoveVolumeCmd removeVolumeCmd = getClient().removeVolumeCmd(volumeName)) {
             removeVolumeCmd.exec();
             log.info("Removed docker volume [name:{}]", volumeName);
             return true;
+        } catch (NotFoundException e) {
+            log.info("No docker volume to remove [name:{}]", volumeName);
```
james-toussaint: `log.error` everywhere*

src/test/java/com/iexec/common/docker/client/AbstractDockerTests.java
```diff
@@ -0,0 +1,51 @@
...
+        DockerClientConfig config = DefaultDockerClientConfig.createDefaultConfigBuilder()
+                .withDockerHost("tcp://localhost:11111")
+                .build();
+        DockerHttpClient httpClient = new ZerodepDockerHttpClient.Builder()
+                .dockerHost(config.getDockerHost())
+                .sslConfig(config.getSSLConfig())
+                .build();
+        DockerClient corruptedDockerClient = DockerClientImpl.getInstance(config, httpClient);
+        DockerClientInstance dockerClientInstance = new DockerClientInstance();
+        ReflectionTestUtils.setField(dockerClientInstance, "client", corruptedDockerClient);
```
james-toussaint, line 48: Wondering if a new constructor with a `DockerClient client` arg wouldn't be cleaner/useful

build.gradle
jbern0rd: It is not about removing the version, it is about using spring BOM and versions defined within the BOM.
As much as possible, we will be using the same set of versions as those defined in a given spring release BOM:
https://github.com/iExecBlockchainComputing/iexec-common/blob/feature/docker-rework-step-1/build.gradle#L40

src/test/java/com/iexec/common/docker/client/AbstractDockerTests.java
jbern0rd: If possible, I prefer not to modify the API during this rework.
I don't find `ReflectionTestUtils` to be wrong in implementation like the current one.
There are other places were it is being used and allows to get the job done.

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
jbern0rd: hmm, I just moved the logging events without modifying the level.
I don't know if we want an `error` here, maybe a `warn` ?
(For me, `error`, should be used to detect things that should not have happened, it is more to detect potential application failures)
I kind of like those descriptions
https://sematext.com/blog/logging-levels/

build.gradle
james-toussaint: I missed that part, thanks!

build.gradle
```diff
@@ -37,6 +37,7 @@ repositories {
 // 'api' configuration allows to expose dependencies with 'compile' scope in pom
 // 'implementation' configuration allows to expose dependencies with 'runtime' scope in pom
 dependencies {
+    implementation(platform("org.springframework.boot:spring-boot-dependencies:2.6.2"))
```
james-toussaint: Very minor, I think convention is
```
implementation platform('org.springframework.boot:spring-boot-dependencies:2.6.2')
testImplementation 'org.springframework:spring-test'
```
(https://docs.gradle.org/current/userguide/platforms.html#sub:bom_import)

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
james-toussaint: We can start with a `warn` yes :+1: 
(We might have this conversation again since we still need to rework return statuses at some point)

build.gradle
jbern0rd: I was wondering about this.
Just to see how it behaves, I tried the kotlin syntax which is supported as well

I can revert back to the groovy documented one.
I am just asking myself if the migration to kotlin scripts is a subject we would like to think about or not

src/main/java/com/iexec/common/docker/client/DockerClientInstance.java
jbern0rd: I agree there is room for discussion here.
Done in f776602

build.gradle
james-toussaint: Not for the next weeks but yes, Kotlin script shall be a topic at some point !

james-toussaint (approved):

src/test/java/com/iexec/common/docker/client/AbstractDockerTests.java
```diff
@@ -0,0 +1,51 @@
...
+import com.github.dockerjava.zerodep.ZerodepDockerHttpClient;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.springframework.test.util.ReflectionTestUtils;
+
+abstract class AbstractDockerTests {
+
+    DockerClientInstance dockerClientInstance = new DockerClientInstance();
+    DockerClientInstance corruptClientInstance = getCorruptInstance();
+
+    DockerClientInstance getCorruptInstance() {
```
mcornaton, line 38: Could you please add a comment to explain why this instance is corrupted?

mcornaton (commented):
Thnaks for this cleanup!
Why did you not apply your pattern on `removeImage` and `removeContainer` (don't test whether the image/container exist but rely on a `NotFoundException`)?
https://github.com/iExecBlockchainComputing/iexec-common/blob/f776602b55cbb38cd7a4a71b86ca093415e0ea42/src/main/java/com/iexec/common/docker/client/DockerClientInstance.java#L349-L368
https://github.com/iExecBlockchainComputing/iexec-common/blob/f776602b55cbb38cd7a4a71b86ca093415e0ea42/src/main/java/com/iexec/common/docker/client/DockerClientInstance.java#L777-L795

build.gradle
jbern0rd: reverted in 5d3a64b

jbern0rd:
> Thnaks for this cleanup! Why did you not apply your pattern on `removeImage` and `removeContainer` (don't test whether the image/container exist but rely on a `NotFoundException`)?
> 
> https://github.com/iExecBlockchainComputing/iexec-common/blob/f776602b55cbb38cd7a4a71b86ca093415e0ea42/src/main/java/com/iexec/common/docker/client/DockerClientInstance.java#L349-L368
> 
> 
> https://github.com/iExecBlockchainComputing/iexec-common/blob/f776602b55cbb38cd7a4a71b86ca093415e0ea42/src/main/java/com/iexec/common/docker/client/DockerClientInstance.java#L777-L795

Because those changes would not have been tested
I focused on network and volume commands in this PR.
There are other PRs to come (look at the PR name :-))

mcornaton (approved):

PR #365, opened 2022-12-15
Feature/docker rework step 2

jbern0rd:


mcornaton (approved):
Thanks for this improvement. A simple suggestion: could we add a tag to all Docker tests such as we could easily exclude them? Maybe `@Tag("Docker")`?

james-toussaint (approved):
Thank you :pray:

PR #373, opened 2023-01-02
Remove unused references

james-toussaint:


mcornaton (approved):

src/test/java/com/iexec/common/docker/client/DockerNetworkTests.java
```diff
@@ -41,7 +41,7 @@ void shouldNotCreateNetworkSinceEmptyName() {
     @Test
     void shouldReturnExistingNetworkIdWenNetworkIsAlreadyPresent() {
         String networkName = getRandomString();
-        String networkId = dockerClientInstance.createNetwork(networkName);
+        dockerClientInstance.createNetwork(networkName);
```
jbern0rd, line 44: I have a rework on this test method in a next PR.
It will be modified again with a check on `networkId` to have a check which makes sense with the test method name.

jbern0rd (approved):

src/test/java/com/iexec/common/docker/client/DockerNetworkTests.java
james-toussaint: Wasn't sure about it, no worry to re-add it later, thanks for the info :ok_hand:

PR #394, opened 2023-03-17
Remove `docker-java` dependencies from `build.gradle`

jbern0rd:


mcornaton (approved):
It seems we can also remove Spring-Test dependency.

jbern0rd:
> It seems we can also remove Spring-Test dependency.

After investigation, the `spring-test` dependency is required for other tests and must be kept for the moment.

PR #397, opened 2023-04-04
Feature/immutable classes in sdk order package

jbern0rd:


src/main/java/com/iexec/common/sdk/order/payload/AppOrder.java
```diff
@@ -16,43 +16,49 @@
...
+    String requesterrestrict;
 
     @Builder
-    public AppOrder(String app, BigInteger price, BigInteger volume, String tag,
-                    String salt, String sign, String datasetrestrict, String workerpoolrestrict,
-                    String requesterrestrict) {
+    AppOrder(
+            String app, BigInteger appprice, BigInteger volume, String tag,
+            String salt, String sign, String datasetrestrict, String workerpoolrestrict,
+            String requesterrestrict) {
```
mcornaton: I usually prefer to have a single parameter per line. I find it easier to read.

mcornaton (approved):
Thanks!

PR #424, opened 2023-08-28
Feature/gradle 8.2.1

jbern0rd:


build.gradle
```diff
@@ -77,6 +77,7 @@ dependencies {
 
     // test
     testImplementation 'org.junit.jupiter:junit-jupiter'
+    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
```
mcornaton, line 80: Should we stay consistent with the other lines?
```suggestion
    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```

jbern0rd: I prefer to stick to gradle release notes
https://docs.gradle.org/current/userguide/upgrading_version_8.html#manually_declaring_dependencies

mcornaton (approved):

PR #426, opened 2023-09-25
Bugfix/remove sgx related code

jbern0rd:


thewhitewizard (approved):

build.gradle
```diff
@@ -50,15 +50,15 @@ dependencies {
 
     implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
 
-    // apache commons.lang3
+    // commons-io
+    implementation 'commons-io:commons-io:2.13.0'
```
mcornaton, line 54: Why do we need it now? Was it included in Maven-Shared-Utils?

jbern0rd: Exactly, please see commit 3fe0059 which contains the explanation

mcornaton (approved):

build.gradle
mcornaton: Great, thanks for the explanation! Let's merge :)

PR #436, opened 2023-12-19
Upgrade to `iexec-commons-poco` 3.2.0

thewhitewizard:


jbern0rd (commented):
Can we keep the empty line in gradle.properties file ?
I kind of like the separation between versions and other properties (I agree it brings more value in files with more properties).

jbern0rd (approved):

PR #451, opened 2024-04-22
Rename `blockchainAdapterUrl` to `configServerUrl` in class `PublicConfiguration`

thewhitewizard:


mcornaton (approved):

src/main/java/com/iexec/common/config/PublicConfiguration.java
```diff
@@ -27,12 +27,13 @@
 public class PublicConfiguration {
     String workerPoolAddress;
     String schedulerPublicAddress;
-    String blockchainAdapterUrl;
+    String configServerUrl;
```
jbern0rd, line 41: Wouldn't we prefer a proper deprecation notice here or at least `@JsonAlias` on the new field name ?

src/main/java/com/iexec/common/config/PublicConfiguration.java
```diff
@@ -21,18 +21,30 @@
 import lombok.Builder;
 import lombok.Value;
 
+/**
+ * The {@code blockchainAdapterUrl} and {@code configServerUrl} fields are repetitive,
+ * so if both are populated, they must contain the same value.
+ * Duplicating these fields ensures worker/scheduler compatibility on different v8 versions.
+ * In v9, the {@code blockchainAdapterUrl} field will be removed.
```
mcornaton: Should we use `@link` tag instead of `@code` to make the doc even clearer?

src/main/java/com/iexec/common/config/PublicConfiguration.java
jbern0rd: Thanks :pray:

jbern0rd (approved):

PR #481, opened 2025-03-19
Make ResultModel immutable and add dealId + taskIndex fields

nabil-Tounarti:


src/main/java/com/iexec/common/result/ResultModel.java
```diff
@@ -18,31 +18,36 @@
...
-@AllArgsConstructor
 @JsonDeserialize(builder = ResultModel.ResultModelBuilder.class)
 public class ResultModel {
 
     //TODO move this to commons-poco
     public static final String EMPTY_CHAIN_ID = Numeric.toHexString(new byte[32]);
     // An ethereum signature is 65 bytes long with R, S and V parts (32 bytes + 32 bytes + 1 byte)
     public static final String EMPTY_WEB3_SIG = Numeric.toHexString(new byte[65]);
+    public static final String EMPTY_DEAL_ID = BytesUtils.EMPTY_HEX_STRING_32;
+    public static final int DEFAULT_TASK_INDEX = BytesUtils.BYTES_32_SIZE;
```
jbern0rd: Are we sure we want to create new constants to mask already existing constants defined as `public static final String` in `BytesUtils`. What is the value of `BytesUtils.BYTES_32_SIZE` ? Are we sure it is a good default value for default task index ?

jbern0rd: To help
`EMPTY_CHAIN_ID = Numeric.toHexString(new byte[32])`
`BytesUtils.EMPTY_HEX_STRING_32 = BytesUtils.bytesToString(new byte[BYTES_32_SIZE]) = Numeric.toHexString(new byte[32]) = EMPTY_CHAIN_ID`

jbern0rd (approved):

